PR #27, opened 2021-05-09
Support &[&str] as CmdArguments

soenkehahn:


justfile
```diff
@@ -7,7 +7,7 @@ test pattern="": build
   cargo test --all -- --test-threads=1 {{ pattern }}
   rm -f 'filename with spaces' foo
 
-test_lib pattern="": build
+test_lib_fast pattern="":
```
casey, line 10: You can use dashes in Just recipe names, which are slightly easier to type.

```suggestion
test-lib-fast pattern="":
```

src/cmd_argument.rs
```diff
@@ -61,6 +61,36 @@ impl CmdArgument for Vec<String> {
...
+/// process as arguments, **without** splitting them by whitespace.
+///
+/// ```
+/// use stir::cmd;
+///
+/// let output: String = cmd!(&["echo", "foo"]);
+/// assert_eq!(output, "foo\n");
+/// ```
+#[rustversion::since(1.51)]
+impl<const N: usize> CmdArgument for &[&str; N] {
```
casey, line 75: Should this be `for [&str; N]`, to avoid the `&`?

src/lib.rs
```diff
@@ -484,6 +484,40 @@ mod tests {
...
+    #[test]
+    fn elements_in_slices_are_not_split_by_whitespace() {
+        in_temporary_directory(|| {
+            let args: &[&str] = &["foo bar"];
+            cmd_unit!("touch", args);
+            assert!(PathBuf::from("foo bar").exists());
+        });
+    }
+
+    #[rustversion::since(1.51)]
```
casey, line 503: Nice, I didn't know about this attribute.

src/cmd_argument.rs
soenkehahn: Oh, right. I didn't realize this was possible. Then the `cmd!(["echo", "foo"])` syntax actually will work.

Do we then want another impl for `&[&str]`? I guess that can also wait for another PR...

justfile
soenkehahn: Yeah, that's better.

PR #30, opened 2021-05-11
Remove trailing newlines from stdout by default

soenkehahn:
This is an experiment to see how it'd look like if captured `stdout` would be trimmed by default. This is a maybe more convenient, but maybe less safe alternative to #32.


soenkehahn:
Closing in favor of #42.

PR #33, opened 2021-05-14
Remove stir::Result

soenkehahn:
Hopefully this would make using `stir` easier, since users can just do `use stir::*;`, without having `stir::Result` shadow `std::result::Result`. And the benefits of having `stir::Result` aren't that big.

There still is `stir::Error`, which can also clash with other error types. I'm inclined to leave that in, at least for now.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,13 +51,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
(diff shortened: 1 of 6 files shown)


casey:
I think it's reasonable. I don't usually like glob imports, but stir has a bunch of random little things that need to be imported (argument wrappers, return value wrappers, etc), so I can see how glob imports would be useful. I personally probably wouldn't glob import things, since I like to know where things come from.

One thing you could do is have a `prelude` module, so users would do `use stir::prelude::*;` This would allow leaving items that are less frequently used, and items that might conflict with other things (like `Result`) out of the prelude.

PR #34, opened 2021-05-14
Allow to capture stdout and stderr for failing child processes

soenkehahn:


soenkehahn:
Incorporated into #37.

PR #35, opened 2021-05-18
[WIP] Fix error behavior

soenkehahn:

PR #37, opened 2021-05-21
Switch to cmd_result!

soenkehahn:
This PR remove the `CmdOutput` impl for `Result` and instead adds a `cmd_result` variant that pins down the return type to `Result<T, Error> where T: CmdOutput`. I'm not sure this should be merged. It's an alternative to #35.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -27,8 +27,7 @@ fn result_succeeding() {
 
     fn test() -> Result<(), Error> {
         // make sure 'ls' is installed
-        let result: Result<(), Error> = cmd_result!(WHICH, "ls");
-        result?;
+        cmd_result!(WHICH, "ls")?;
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


tests/integration.rs
```diff
@@ -27,8 +27,7 @@ fn result_succeeding() {
 
     fn test() -> Result<(), Error> {
         // make sure 'ls' is installed
-        let result: Result<(), Error> = cmd!(WHICH, "ls");
-        result?;
+        cmd_result!(WHICH, "ls")?;
```
soenkehahn, line 30: I'm really surprised that this works without type annotations. I thought it would have to be e.g. `let () = cmd_result!(WHICH, "ls")?;`. There seems to be some type defaulting to be going on.

(cc: @casey)

soenkehahn: Here's some explanation of the rustc behavior: https://www.reddit.com/r/rust/comments/f5nw1o/why_does_rust_infer_when_the_generic_type_cannot/

casey: That's interesting, I guess `()` is just a default in some circumstances.

PR #40, opened 2021-05-25
Allow to control the current directory of child processes

soenkehahn:


src/cmd_argument.rs
```diff
@@ -121,3 +122,33 @@ impl CmdArgument for LogCommand {
         config.log_command = true;
     }
 }
+
+/// Please, see the [`CmdArgument`] implementation for [`CurrentDir`] below.
```
casey, line 126: I think you can leave out the "Please," it sounds a little desperate 😜

src/cmd_argument.rs
```diff
@@ -121,3 +122,33 @@ impl CmdArgument for LogCommand {
...
+pub struct CurrentDir<T: AsRef<Path>>(pub T);
+
+/// By default child processes inherit the current directory from their
+/// parent. You can overwrite this with [`CurrentDir`]:
```
casey, line 130: Maybe "override" instead of "overwrite"

src/cmd_argument.rs
```diff
@@ -121,3 +122,33 @@ impl CmdArgument for LogCommand {
...
+///
+/// ```
+/// # #[cfg(target_os = "linux")]
+/// # fn test() {
+/// use stir::*;
+///
+/// let output: String = cmd!("pwd", CurrentDir("/tmp"));
+/// assert_eq!(output, "/tmp\n");
+/// # }
+/// # #[cfg(not(target_os = "linux"))]
```
casey, line 140: Is this necessary to compile? Could you do:

````rust
# fn test() {
# #[cfg(unix)]
# {
// test code here
# }
# }
```

Also note `#[cfg(unix)]`, since you aren't using any Linux functionality, and you would want the test to run on e.g. a BSD.

soenkehahn: Well, this *compiles* fine on windows, mac & linux. The assertion is the problem, though. On github's ci on windows `output` would be `"/d\n"` and on mac it'd be `"/private/tmp\n"`, IIRC. So the assertion would fail. That's why the test is only enabled on `linux`.

But we could solve this differently:

- Change the directory from `/tmp` to `/` and then run the tests on `cfg(unix)`. I think that would work on mac and linux. I do think for documentation `/tmp` is slightly nicer.
- Provide different assertions for mac, linux and windows, but hide the assertions for mac and windows.
- Some other idea?

I think as long as we don't have ci for bsd, I'd consider it as unsupported.

src/cmd_argument.rs
soenkehahn: https://github.com/soenkehahn/stir/pull/40/commits/007553105ef574bce559446631d7686aebc7b86d

src/cmd_argument.rs
casey: The thing that I was wondering about was just the empty test on `not(linux)`:

```rust
/// # #[cfg(not(target_os = "linux"))]
/// # fn test() {}
/// # test();
```

Does that need to be there at all?

> I think as long as we don't have ci for bsd, I'd consider it as unsupported.

I see your point, but BSDs (and other Unixes) are so close to Linux (aside from system-specific syscalls, which are pretty uncommonly used.) that I think it's worth considering your target as being Unix, instead of considering non-Linux OSes as being unsupported.

For example, I think Just works fine on BSD (and it's made its way into the FreeBSD package manager, so there are a few users at least) and I don't believe I've ever gotten a BSD-specific bug, nor have I ever had to write any BSD-specific code.

casey:
"Eradicate all occurrences of common courtesy from the documentation"


LMAAAAAAAAAAAAAAO

src/cmd_argument.rs
soenkehahn: So `test()` was being executed on all platforms. But then for non-linux it was just an empty function. But yeah, we can do that simpler, just pushed a commit: https://github.com/soenkehahn/stir/pull/40/commits/33ffbe0eb2dbd4ef574608f76d4599cc96b4073b

soenkehahn: Given that we do have another test that makes sure that `CurrentDir` works on all platforms, I don't think we need to execute the assertion of this doctest everywhere.

PR #41, opened 2021-05-27
Switch to newtype wrapper `Stdout` to capture stdout (instead of `String`)

soenkehahn:


src/cmd_argument.rs
```diff
@@ -69,7 +69,7 @@ impl CmdArgument for Vec<String> {
 /// ```
 /// use stir::*;
 ///
-/// let output: String = cmd!(["echo", "foo"]);
+/// let Stdout(output) = cmd!(["echo", "foo"]);
```
soenkehahn, line 72: Users of `stir` have to provide some assistance to the compiler to guide type inference to figure out the return type of `cmd!` (and friends). I wonder whether it's nicer to pattern-match with `Stdout` than to have to provide type annotations for that.

soenkehahn:
Closing in favor of #42.

PR #42, opened 2021-05-28
Trimmed and untrimmed stdout

soenkehahn:
This PR removes the `CmdOutput` impl for `String` and instead introduces 2 new-type wrappers around `String`: `StdoutUntrimmed` and `StdoutTrimmed`. They can be used like this:

```rust
use stir::*;

let StdoutTrimmed(ls_path) = cmd!("which ls");
let StdoutUntrimmed(output) = cmd!("cat some_file");
```

This is slightly less convenient than having a `String` impl, but it forces users to make a conscious decision about whether they want trimmed or untrimmed output. In my experience it happens often that I capture the `stdout` of a process, and try to do something with it, only to realize later that my program fails, because the output contains a trailing `\n`. This PR would help with that and probably save some time. 

For example this program type-checks with `stir` **before** this PR:

```rust
use stir::*;

fn main() {
  // start a container
  let container_id: String = cmd!("docker create ubuntu:latest");

  // do something with the container
  cmd_unit!("docker cp", format!("{}:/etc/lsb-release", container_id), "ubuntu-latest-lsb-release");
  cmd_unit!("cat ubuntu-latest-lsb-release");

  // shut down the container
  cmd_unit!("docker rm", container_id);
}
```

But it fails at runtime with this rather confusing error message:
```
"docker cp" requires exactly 2 arguments.
See 'docker cp --help'.

Usage:  docker cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-
	docker cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH

Copy files/folders between a container and the local filesystem
thread 'main' panicked at 'cmd!: docker cp 81c66997c246a24cabb777eabe0d9c82e80769848998da94eb1f965dc2eaaa95 :/etc/lsb-release ubuntu-latest-lsb-release:
  exited with exit code: 1', /home/shahn/stir/src/error.rs:36:23
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The reason is, that `container_id` contains the container id *and* a trailing `\n`. This PR hopefully means that users will not run into these errors anymore.

### DIFF
--- a/src/cmd_output.rs
+++ b/src/cmd_output.rs
@@ -64,11 +64,11 @@ pub struct StdoutTrimmed(pub String);
 /// ```
 impl CmdOutput for StdoutTrimmed {
     fn prepare_config(config: &mut Config) {
-        <StdoutUntrimmed as CmdOutput>::prepare_config(config);
+        StdoutUntrimmed::prepare_config(config);
     }
 
     fn from_run_result(config: &Config, result: Result<RunResult, Error>) -> Result<Self, Error> {
-        let StdoutUntrimmed(stdout) = CmdOutput::from_run_result(config, result)?;
+        let StdoutUntrimmed(stdout) = StdoutUntrimmed::from_run_result(config, result)?;
         Ok(StdoutTrimmed(stdout.trim().to_owned()))
     }
 }
--- a/src/cmd_output.rs
+++ b/src/cmd_output.rs
@@ -38,14 +38,25 @@ impl CmdOutput for () {
     }
 }
 
-/// Please, see the [`CmdOutput`] implementation for [`Stdout`] below.
+/// See the [`CmdOutput`] implementation for [`Stdout`] below.
 #[derive(Debug, PartialEq, Clone)]
 pub struct Stdout(pub String);
 
 /// Returns what the child process writes to `stdout`, interpreted as utf-8,
 /// collected into a string. This also suppresses output of the child's `stdout`
-/// to the parent's `stdout`. (Which would be the default when not using [`String`]
+/// to the parent's `stdout`. (Which would be the default when not using [`Stdout`]
 /// as the return value.)
+///
+/// It's recommended to pattern-match to get to the inner [`String`].
+/// This will make sure that the return type can be inferred.
+/// Here's an example:
+///
+/// ```
+/// use stir::*;
+///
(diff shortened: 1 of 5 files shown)


src/cmd_output.rs
```diff
@@ -39,11 +38,54 @@ impl CmdOutput for () {
...
+///
+/// # #[cfg(unix)]
+/// # {
+/// let StdoutTrimmed(output) = cmd!("which ls");
+/// assert!(Path::new(&output).exists());
+/// # }
+/// ```
+impl CmdOutput for StdoutTrimmed {
+    fn prepare_config(config: &mut Config) {
+        <StdoutUntrimmed as CmdOutput>::prepare_config(config);
```
casey, line 67: Why doesn't this recurse?

casey:
The seems like a good change to me. I think I would probably leave the implementation for `String`, but I see the logic for leaving it out, and if users want it, it can always be added back as a backwards compatible change.

src/cmd_output.rs
soenkehahn: The implementation for `StdoutTrimmed` invokes the implementation for `StdoutUntrimmed`. Just to make it DRY. Do you it'd be clearer to just duplicate the implementation? Same question for `from_run_result`...

src/cmd_output.rs
```diff
@@ -52,9 +94,11 @@ impl CmdOutput for String {
     #[doc(hidden)]
     fn from_run_result(config: &Config, result: Result<RunResult, Error>) -> Result<Self, Error> {
         let result = result?;
-        String::from_utf8(result.stdout).map_err(|_| Error::InvalidUtf8ToStdout {
-            full_command: config.full_command(),
-        })
+        Ok(StdoutUntrimmed(String::from_utf8(result.stdout).map_err(
+            |_| Error::InvalidUtf8ToStdout {
```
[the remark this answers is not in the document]
soenkehahn: That's a good point. I'll do that in a different PR though.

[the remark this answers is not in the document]
soenkehahn: https://github.com/soenkehahn/stir/issues/43

src/cmd_output.rs
soenkehahn: Tweaked this a little after offline consultation.

PR #46, opened 2021-06-02
Avoid splitting by whitespace by default

soenkehahn:


soenkehahn:
This PR removes the behavior to split commands by whitespace by default. Instead it introduces two now ways to do that:

1. A `Split` newtype wrapper around `&str`.
2. Allowing to use `str::split`, `str::split_whitespace` or `str::split_acsii_whitespace`.

I think I'd like to only add one of the two options. Any opinions, which one?

Also, I'm not super happy with this overall. I think that e.g. `cmd!("command --foo", some_argument, "--some-config true");` is just much better than `cmd!("command --foo".split(' '), some_argument, "--some-config true".split(' '));`. :/

casey:
> This PR removes the behavior to split commands by whitespace by default. Instead it introduces two now ways to do that:
> 
> 1. A `Split` newtype wrapper around `&str`.
> 2. Allowing to use `str::split`, `str::split_whitespace` or `str::split_acsii_whitespace`.
> 
> I think I'd like to only add one of the two options. Any opinions, which one?

I think I like `Split` a little bit more, since it's consistent with everything else in the library. However, `.split` can be used with different delimiters, which might be useful, and it doesn't require learning a new thing. I'm torn. I would probably add both, since I don't think there's a huge downside, but if I had to only add one it would be `.split`.

> Also, I'm not super happy with this overall. I think that e.g. `cmd!("command --foo", some_argument, "--some-config true");` is just much better than `cmd!("command --foo".split(' '), some_argument, "--some-config true".split(' '));`. :/

You could also write it like this:

```rust
cmd!("command", "--foo", some_argument, "--some-config", "true");
```

I agree that this is a little worse, but I suspect that string splitting bugs will be inevitable otherwise, and will be difficult to track down, and crop up unexpectedly. For example, a utility is only tested with paths without spaces, but a user then uses it with paths with spaces, triggering splitting and causing bugs.

casey:
An additional way that users can avoid needing to split strings is by using `=` to separate options and values, if the command supports it:

```rust
cmd!("command", "--foo", some_argument, "--some-config=true");
```

soenkehahn:
> I agree that this is a little worse, but I suspect that string splitting bugs will be inevitable otherwise, and will be difficult to track down, and crop up unexpectedly. For example, a utility is only tested with paths without spaces, but a user then uses it with paths with spaces, triggering splitting and causing bugs.

Yeah, I completely agree. This is especially bad in a language like rust where programmers rely on the type system to catch bugs like that.

I'm pretty convinced we should merge this.

casey:
I think this looks good!

casey:
I had some thoughts about the clones, but I think those are really only performance concerns, so I'll open an issue.

src/lib.rs
```diff
@@ -25,51 +25,34 @@
...
-//! # }
-//! # #[rustversion::before(1.51)]
-//! # fn test() {}
-//! # test();
+//! let StdoutTrimmed(_) = cmd!("echo foo");
 //! ```
 //!
-//! Before rust version `1.51`, instead of arrays, use [`Vec<&str>`]:
+//! In this code `cradle` tries to run a process from an executable called
+//! `"echo foo"`, including the space in the executable file name.
```
soenkehahn, line 42:
```suggestion
//! `"echo foo"`, including the space in the file name of the executable.
```

PR #47, opened 2021-06-02
Add `source` fields to error variants for invalid utf-8

### LINKED ISSUE #43: Include source error for InvalidUtf8ToStdout & InvalidUtf8ToStderr

soenkehahn:
Fixes #43.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -14,11 +14,11 @@ pub enum Error {
     },
     InvalidUtf8ToStdout {
         full_command: String,
-        source: FromUtf8Error,
+        source: Arc<FromUtf8Error>,
     },
     InvalidUtf8ToStderr {
         full_command: String,
-        source: FromUtf8Error,
+        source: Arc<FromUtf8Error>,
     },
 }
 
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -60,10 +62,34 @@ impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match self {
             Error::CommandIoError { source, .. } => Some(&**source),
-            Error::NoArgumentsGiven
-            | Error::NonZeroExitCode { .. }
-            | Error::InvalidUtf8ToStdout { .. }
-            | Error::InvalidUtf8ToStderr { .. } => None,
+            Error::InvalidUtf8ToStdout { source, .. }
+            | Error::InvalidUtf8ToStderr { source, .. } => Some(&**source),
```
casey, line 66: Is it not possible to produce a `&dyn` reference without an arc?

casey:
I personally don't like the untyped `source` approach, since I think it's useful to have access to the concrete error type, and useful documentation to see what can cause an error. However, aside from that, this looks good!

src/error.rs
soenkehahn: Good catch. I thought I needed the `Arc` because `cradle::Error` implements `Clone`. But `FromUtf8Error` does too.

soenkehahn: Never mind. On older rust versions `FromUtf8Error` does *not* implement `Clone`. So I'll put the `Arc` back in.

PR #50, opened 2021-06-03
Implement syntactic shortcut for `Split`

soenkehahn:
(This PR depends on #46, so #46 should be merged first.)

This PR implements a shorthand `~` for the newtype wrapper `Split`. So instead of `cmd_unit!(Split("mkdir -p foo"));` you could do `cmd_unit!(~"mkdir -p foo");`. As this is a shorthand for the `Split` constructor, it would only work for values of type `&str`.

This is a departure from the stance that `cmd!` and friends don't introduce any new syntax and only use their macro-powers to allow a variable number of arguments. However -- given that any other solution to whitespace splitting feels unsatisfying (as discussed in #46) -- I think this might be the least bad design.

One open question is, whether this negatively impacts code formatting.

Alternatively, I think we could further restrict the `~` shorthand syntax to only work in front of `&str` literals. But I don't really see how it's harmful to do:
```rust
let command = "echo foo";
cmd_unit!(~command);
```

### DIFF
--- a/src/cmd_argument.rs
+++ b/src/cmd_argument.rs
@@ -11,6 +11,7 @@ impl<T> CmdArgument for &T
 where
     T: CmdArgument + Clone,
 {
+    #[doc(hidden)]
     fn prepare_config(self, config: &mut Config) {
         <T as CmdArgument>::prepare_config(self.clone(), config);
     }
@@ -48,35 +49,24 @@ impl<'a> CmdArgument for Split<'a> {
     }
 }
 
-/// All elements of the given [`Vec`] are being passed into the child
-/// process as arguments, **without** splitting them by whitespace.
-///
-/// This can come in handy to avoid whitespace splitting, even if you only want
-/// to encode a single argument:
+/// All elements of the given [`Vec`] are used as arguments to [`cmd!`].
+/// Same as passing in the elements separately.
 ///
 /// ```
-/// use std::path::PathBuf;
 /// use cradle::*;
 ///
-/// cmd_unit!("touch", vec!["filename with spaces"]);
-/// assert!(PathBuf::from("filename with spaces").exists());
+/// let StdoutTrimmed(output) = cmd!(vec!["echo", "foo"]);
+/// assert_eq!(output, "foo");
 /// ```
-impl CmdArgument for Vec<&str> {
+impl<T> CmdArgument for Vec<T>
+where
+    T: CmdArgument,
+{
     #[doc(hidden)]
     fn prepare_config(self, config: &mut Config) {
-        for argument in self {
-            config.arguments.push(argument.to_string());
+        for t in self.into_iter() {
+            <T as CmdArgument>::prepare_config(t, config);
         }
-    }
-}
-
-/// Similar to the implementation for [`Vec<&str>`].
-/// All elements of the given [`Vec`] are being passed into the child
-/// process as arguments, **without** splitting them by whitespace.
-impl CmdArgument for Vec<String> {
-    #[doc(hidden)]
-    fn prepare_config(self, config: &mut Config) {
-        config.arguments.extend(self);
     }
 }
 
--- a/src/cmd_argument.rs
+++ b/src/cmd_argument.rs
@@ -49,7 +49,7 @@ pub struct Split<'a>(pub &'a str);
 /// assert_eq!(output, "foo");
 /// ```
 ///
-/// Since this is such a common case, `cradle` also provides a syntactical shortcut
+/// Since this is such a common case, `cradle` also provides a syntactic shortcut
 /// for [`Split`]: the `@` symbol:
 ///
 /// ```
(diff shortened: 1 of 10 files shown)


soenkehahn:
Oh, and is there a better symbol to use than `~`?

casey:
I actually think that needing to call `split` isn't that bad, and I've started to see the benefit of the macro not doing anything weird.

That being said, I think that `~` might be confusing, since at a glance, it looks like it might have something to do with home directory expansion.

I made [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a0af73c0eca491df4b2e40a6debed4e5) to check for all the symbols that could be used. 

Of these, I lean towards `@`, since it's already used for array operations in Bash.

src/lib.rs
```diff
@@ -25,52 +25,44 @@
...
 //!
 //! ```
-//! use std::path::PathBuf;
 //! use cradle::*;
 //!
-//! let StdoutTrimmed(_) = cmd!("touch", vec!["filename with spaces"]);
-//! assert!(PathBuf::from("filename with spaces").exists());
+//! let StdoutTrimmed(output) = cmd!(~"echo foo");
+//! assert_eq!(output, "foo");
 //! ```
```
soenkehahn, line 64: Here's some documentation about how `Split` and the `~` shortcut relate.

soenkehahn:
Yes, `~` can be confused with the home directory shortcut in shells. `@` is better. Not sure whether there's other even better options. We could also consider something that's not just a prefix, like `cmd!({"echo foo"});`. Or `cmd!(|"echo foo"|);`.

soenkehahn:
One consequence of having this magic syntax, is that it only works at the top-level. So e.g. this doesn't work: `cmd!([~"echo foo"]);`. It's not clear to me whether having this work would ever be useful, but it'd be nice for consistency. It is easy to work around the issue with `cmd!([Split("echo foo")]);`, of course.

src/cmd_argument.rs
```diff
@@ -49,6 +49,16 @@ pub struct Split<'a>(pub &'a str);
 /// assert_eq!(output, "foo");
 /// ```
 ///
+/// Since this is such a common case, `cradle` also provides a syntactical shortcut
+/// for [`Split`]: the `@` symbol:
```
casey, line 53:
```suggestion
/// for [`Split`], the `@` symbol:
```

src/lib.rs
```diff
@@ -54,6 +54,16 @@
 //! split it by whitespace and then use the resulting words as if they were passed
 //! into [`cmd!`] as separate arguments.
 //!
+//! And -- since this is such a common case -- `cradle` provides a syntactical shortcut
+//! for [`Split`]: the `@` symbol:
```
casey, line 58:
```suggestion
//! And -- since this is such a common case -- `cradle` provides a syntactic shortcut
//! for [`Split`], the `@` symbol:
```

src/lib.rs
```diff
@@ -179,36 +189,55 @@ use std::{
...
         let context = $crate::Context::production();
-        $crate::cmd_result_with_context!(context, $($args),+)
+        $crate::cmd_result_with_context!(context, $($args)*)
     }}
 }
 
+#[doc(hidden)]
+#[macro_export]
+macro_rules! prepare_config {
+    ($config:ident # @ $head:expr, $($tail:tt)*) => {
```
casey, line 219: One thing I thought of to make this clearer would be to use `config: …, args: …` instead of `#` to document what's going on. Like so:

```rust
    (config: $config:ident, args: @ $head:expr, $($tail:tt)*) => {
```

This would make it a little clearer what's happening, although would be a little longer.

casey (commented):
> One consequence of having this magic syntax, is that it only works at the top-level.

I think that's okay. It would be nice if it worked everywhere, but it would make the macro insane. I think you'd have to add an expression parser.

This looks good to me!

src/lib.rs
soenkehahn: Yeah, I really like that! I was trying to make the macro invocation syntax a bit clearer by introducing the `#`, but wasn't really happy with that. Didn't occur to me to use named arguments.

soenkehahn:
> That being said, I think that `~` might be confusing, since at a glance, it looks like it might have something to do with home directory expansion.
> 
> I made [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a0af73c0eca491df4b2e40a6debed4e5) to check for all the symbols that could be used.
> 
> Of these, I lean towards `@`, since it's already used for array operations in Bash.

Isn't that an argument for *not* using `@` though? Since we don't want to summon associations with different behavior in `bash`. (As far as I know `@` doesn't do anything related to word splitting, does it?)

What do you think of `%`? (If you squint at the symbol, it looks like a line splitting two things. Although I also like `%` apart from that.)

PR #59, opened 2021-06-18
Add Stdin as CmdArgument

soenkehahn:


src/collected_output.rs
```diff
@@ -1,28 +1,38 @@
...
         mut child_stdout: ChildStdout,
         mut child_stderr: ChildStderr,
     ) -> Self
     where
         Stdout: Write + Send + Clone + 'static,
         Stderr: Write + Send + Clone + 'static,
     {
+        let mut config_stdin = Vec::new();
+        std::mem::swap(&mut config.stdin, &mut config_stdin);
+        let stdin_join_handle = thread::spawn(move || {
```
casey, line 28: Does this use a background thread to avoid deadlocking?

src/cmd_argument.rs
```diff
@@ -268,3 +268,24 @@ impl CmdArgument for &Path {
         self.to_path_buf().prepare_config(config);
     }
 }
+
+/// See the [`CmdArgument`] implementation for [`Stdin`] below.
+pub struct Stdin<'a>(pub &'a str);
```
casey, line 273: I think the most general version of this would be:

```suggestion
pub struct Stdin<T: Into<String>>(pub T);
```

You could do:

```
pub struct Stdin<T: AsRef<str>>(pub T);
```

But with `T: Into<String>` you can avoid a clone when a `String` is passed.

src/cmd_argument.rs
```diff
@@ -268,3 +268,24 @@ impl CmdArgument for &Path {
...
+
+/// Passes the given [`&str`] into the child on `stdin`.
```
casey, line 275: Not sure how to word this, but something that makes it clear that `Stdin` can be used multiple times might be good:

```suggestion
/// Writes the given [`&str`] to the child's standard input.  If `Stdin` is used multiple times,
/// each string will be written written to the child's standard input in the order it appears.
```

src/collected_output.rs
```diff
@@ -55,12 +65,16 @@ impl Waiter {
...
+            stdin: stdin_join_handle,
             stdout: stdout_join_handle,
             stderr: stderr_join_handle,
         }
     }
 
     pub(crate) fn join(self) -> io::Result<CollectedOutput> {
+        self.stdin
+            .join()
+            .expect("stdout relaying thread panicked")?;
```
casey, line 77: Can this return an `io::Error` instead of panicking?

If the child process doesn't read from stdin, would this cause the `.expect` to panic?

soenkehahn: > Can this return an `io::Error` instead of panicking?

No. Well, it's a bit complicated. There's two layers of `Result`s here:
The call to `join` (in this context here) returns `std::thread::Result<io::Result<()>>`.
The outer `thread::Result` layer is introduced by [`JoinHandle::join`](https://doc.rust-lang.org/std/thread/struct.JoinHandle.html#method.join) (the one that's being called here).
That `join` returns an `Err` if and only if the child thread panics.
The error type in there is `Box<dyn Any + Send + 'static>` and I don't think that can be converted to an `io::Error`.
Also, in the case where the thread panics, I think it's best to re-throw that panic in the main thread.
Instead of just letting the main thread continue, even though something went terribly wrong.

The inner `Result` layer here is the `io::Result` that is being produced by the closure passed to `std::thread::spawn`.
And that `Result` is *not* converted into a panic, but errors are just propagated normally. (As in `Waiter::join` returns an `io::Result`.)

Hope this explanation helps and that's what you were asking.

> If the child process doesn't read from stdin, would this cause the `.expect` to panic?

No. If the child process doesn't read from stdin, and the parent writes to it, then it'll probably go into the OS buffer (I think) and the writing thread will exit normally. If that buffer is full, one of the `write` calls in the writing thread will block. And then the `join` call would hang. :scream: I'll have to test that. Good catch...

src/cmd_argument.rs
soenkehahn: What do you think about [this](https://github.com/soenkehahn/cradle/pull/59/commits/31592105af5fe5eea574a5a0287f5eeaa65a9aac#diff-314b0e8d3543865f2dac2e6e8ae1d3cc4abb8dadebf2a4576d8674c07c010ea4R275-R277)?

src/collected_output.rs
soenkehahn: Yes, exactly. This has to be executed concurrently with the main thread that will spawn the child process. But this also has to be executed concurrently with the other threads in here, that collect `stderr` and `stdout`. Since a child process could potentially do things like write to `stdout` only after having received something on `stdin` (which is pretty normal unix filter behavior) or read from `stdin` only after being able to write to `stdout` (which is pretty weird, but possible). There's also OS buffers involved, but that of course only helps with deadlocks as long as they aren't full.

Of course I wish there were a more elegant and more light-weight way to deal with this than OS threads. But I don't see an obvious solution that is easy to implement and avoids threads.

src/cmd_argument.rs
soenkehahn: Nice, I'll use `Into<String>`.

src/collected_output.rs
soenkehahn: I added return types to the closures for the three threads in here in hope to make it a bit clearer what's going on with the error handling.

Also, I pushed [the `stdin-experiment` branch](https://github.com/soenkehahn/cradle/compare/063135974db11e0f69f132595f64f4dc0b94d507...stdin-experiment) to test the behavior when
- the parent writes a lot of bytes into the child's `stdin`,
- the child terminates without reading from `stdin`.

I suspected that the `write` call would hang indefinitely in the writing thread, causing this `JoinHandle::join` call to hang. But that's not what happens. Instead when the child terminates, the read end of the pipe gets closed and the blocking write call errors with an `io::Error`. (That error will then get propagated properly, meaning `Waiter::join` returns an `Err`. So i.e. using `cmd_result!` will allow users to handle that error through `Result`.)

I think that that might be exactly the behavior we want. If users try to write a bunch of stuff into the child's `stdin`, but that process never reads from it and instead closes `stdin` (for example by terminating), it seems fair to throw an error. It's a bit weird that there's the OS buffer in between, which means that if the parent writes less bytes to the child's `stdin`, the error never shows up. But there's little we can do about that. I think this is just how pipes work. Or is there a different, more reasonable behavior?

In any case we should add a test for documenting whatever the behavior is.

(I also did a similar experiment where the parent fills the buffer of the child's `stdin`, the child never reads from it, but it also never terminates. In that case the writing thread just hangs on one of the `write` calls. And the main thread hangs on waiting for the child process to terminate. Which seems like correct behavior.)

soenkehahn: I now added this test: https://github.com/soenkehahn/cradle/pull/59/commits/bff35517d01862e680b4f2b203b1cbb51eb8aad2

src/collected_output.rs
casey: Makes sense. I think this is totally fine. After all, a process is being launched, which is heavier weight than a single thread, so a process + a thread won't be much slower than just a process.

src/collected_output.rs
casey: Gotcha, that all makes sense. Correct me if I'm wrong, but I think we now don't think that any of the worker threads can panic?

If so, then it's reasonable to leave the `.expect` in, since we think it can't be triggered.

Another option, not sure how valuable it is, would be to map the err value of `std::thread::Result` to an `io::Error`, using `std::io::Error::new(std::io::ErrorKind::Other, join_error)`.

This would allow propagating the join error, if it were to occur, as an io::Error.

src/cmd_argument.rs
casey: SGTM

casey: SGTM

src/collected_output.rs
casey: > It's a bit weird that there's the OS buffer in between, which means that if the parent writes less bytes to the child's stdin, the error never shows up. But there's little we can do about that. I think this is just how pipes work. Or is there a different, more reasonable behavior?

I missed this question. This seems fine to me. The user requested that we write some bytes to the child's stdin, but the child didn't read them all, that seems okay, since there are lots of circumstances in which a command might not read all of its input.

soenkehahn: > Gotcha, that all makes sense. Correct me if I'm wrong, but I think we now don't think that any of the worker threads can panic?

Yes, exactly.

src/config.rs
```diff
@@ -1,10 +1,11 @@
-use std::{ffi::OsString, path::PathBuf};
+use std::{ffi::OsString, path::PathBuf, sync::Arc};
 
 #[doc(hidden)]
 pub struct Config {
     pub(crate) arguments: Vec<OsString>,
     pub(crate) log_command: bool,
     pub(crate) working_directory: Option<PathBuf>,
+    pub(crate) stdin: Arc<Vec<String>>,
```
soenkehahn, line 8: After #67 I had to either pass in a `&mut Config` into `Waiter::spawn_standard_stream_relaying` (which I think is weird) or make this an `Arc`. I think this is fine. cc:  @casey

soenkehahn: I think having an `Arc` here is actually better than the solution I had before merging in #67. In that state I was mutating the passed in `Config` to avoid cloning of `stdin`.

casey: I think the `Arc` is good.

casey: You could also use `std::mem::swap` to move out of the config. It is a little weird to mutate the config though.

soenkehahn: Yeah, that's what I had before #67. But I think it's better to have an `Arc` than to have mutability.

PR #60, opened 2021-06-18
Allow &Path and PathBuf as CmdArguments

soenkehahn:


src/cmd_argument.rs
```diff
@@ -232,3 +232,33 @@ where
...
+    #[doc(hidden)]
+    fn prepare_config(self, config: &mut Config) {
+        config.arguments.push(self.into_os_string());
+    }
+}
+
+/// Arguments of type [`&Path`] are passed into the child process
+/// as arguments.
+///
+/// Similar to the impl for [`PathBuf`].
```
soenkehahn, line 256: Should this also have a doctest example?

src/lib.rs
```diff
@@ -935,6 +936,24 @@ mod tests {
                 .unwrap();
             assert_eq!(context.stderr(), "+ echo 'foo bar'\n");
         }
+
+        #[test]
+        #[cfg(unix)]
+        fn arguments_with_invalid_utf8_will_be_logged_with_lossy_conversion() {
```
soenkehahn, line 942: I added this test, since now it's possible to create command arguments with invalid utf8.

src/lib.rs
```diff
@@ -1216,4 +1235,63 @@ mod tests {
...
+                let StdoutUntrimmed(output) = cmd!("cat", file);
+                assert_eq!(output, "test-contents");
+            })
+        }
+
+        #[test]
+        fn ref_path_as_executable() {
+            in_temporary_directory(|| {
+                let file: &Path = &write_test_script();
+                let StdoutTrimmed(output) = cmd!(file);
```
soenkehahn, line 1273: One weird detail with `Path` and `PathBuf` is this: If users create a path

- without slashes,
- that doesn't point to an existing file,
- they try to use that path as the executable in `cmd!`, and
- an executable with the same name can be found in one of the directories in `$PATH`,

then that executable from `$PATH` will be executed. This is a bit weird, since values of type `Path` or `PathBuf` are not supposed to point to files from `$PATH` that way. (As evidenced by `.exists()` returning `false`.) However `Command::new` behaves the same way, so I think this is weird, but ok.

soenkehahn:
cc: @casey

src/cmd_argument.rs
```diff
@@ -232,3 +232,33 @@ where
         config.working_directory = Some(self.0.as_ref().to_owned());
     }
 }
+
+/// Arguments of type [`PathBuf`] are passed into the child process
```
casey, line 236:
```suggestion
/// Arguments of type [`PathBuf`] are passed to the child process
```

src/cmd_argument.rs
```diff
@@ -232,3 +232,33 @@ where
...
+/// use cradle::*;
+/// use std::path::PathBuf;
+///
+/// let current_dir: PathBuf = std::env::current_dir().unwrap();
+/// cmd_unit!("ls", current_dir);
+/// ```
+impl CmdArgument for PathBuf {
+    #[doc(hidden)]
+    fn prepare_config(self, config: &mut Config) {
+        config.arguments.push(self.into_os_string());
```
casey, line 249: I didn't actually try this, but suggesting it for consistency.

```suggestion
        config.arguments.push(self.into());
```

src/cmd_argument.rs
```diff
@@ -232,3 +232,33 @@ where
...
+    }
+}
+
+/// Arguments of type [`&Path`] are passed into the child process
```
casey, line 253:
```suggestion
/// Arguments of type [`&Path`] are passed to the child process
```

src/cmd_argument.rs
casey: Yeah, maybe it should. It's really simple, but if the user happens to land on that doc item without knowing the library, a doctest example would be helpful.

src/lib.rs
casey: I think that's okay. I agree that it's conceptually weird, but you can also think of, for example `Path::new("echo")` as being a path to the `echo` binary, which is relative to one of the paths in `$PATH`.

casey (commented):
Sweet, some minor comments but overall looks good to me. Getting close to being featureful enough to shill it to users!

src/cmd_argument.rs
soenkehahn: Done in https://github.com/soenkehahn/cradle/pull/60/commits/1d5014d50b1e1e35fe0db03eeebf18610bdd6a16.

src/cmd_argument.rs
soenkehahn: Yup, works.

src/cmd_argument.rs
soenkehahn: Done.

PR #64, opened 2021-06-25
Make `Split` take anything that implements `AsRef<str>`

casey:


soenkehahn:
Good idea. This would allow e.g. `String`, right? Also, why not `Into<String>`?

src/cmd_argument.rs
```diff
@@ -37,7 +37,7 @@ impl CmdArgument for String {
 }
 
 /// See the [`CmdArgument`] implementation for [`Split`] below.
-pub struct Split<'a>(pub &'a str);
+pub struct Split<T: AsRef<str>>(pub T);
```
soenkehahn, line 40: Why not use `Into<String>`?

src/cmd_argument.rs
```diff
@@ -47,6 +47,9 @@ pub struct Split<'a>(pub &'a str);
 ///
 /// let StdoutTrimmed(output) = cmd!(Split("echo foo"));
 /// assert_eq!(output, "foo");
+///
+/// let StdoutTrimmed(output) = cmd!(Split(format!("echo {}", 100)));
+/// assert_eq!(output, "100");
```
soenkehahn, line 52: Maybe we should also have a test in `lib.rs`? So far I'm using doctests in this project more for (automatically checked) documentation, and not for test coverage.

soenkehahn (commented):
Thanks a lot!

src/cmd_argument.rs
casey: If the argument is of type `&str`, `Into<String>` would be a clone, which isn't necessary, since `split_whitespace` takes `&self`.

src/cmd_argument.rs
casey: Sounds good! I'll add a real test.

casey:
Just added a test and merged in latest master.

src/cmd_argument.rs
soenkehahn: Right, thanks for explaining.

src/cmd_argument.rs
soenkehahn: Thank you!

soenkehahn:
Thank you!!!

PR #67, opened 2021-06-25
Avoid cloning Config

casey:
This avoids a couple clones of the `Config` struct. Definitely a micro optimization, but the code is a little nicer too.


soenkehahn:
Nice, thanks!

PR #71, opened 2021-06-27
Add CmdArgument impls for OsString and OsStr

casey:
Allows using `OsString` and `&OsStr` as arguments to `cmd`. Among other things, OsString is the type of environment variables and arguments, and `&OsStr` is the type of `Path::file_name`. This diff also changes all `CmdArgument` impls to bottom out in the impl for `OsString`, because that's the actual type of `Config::arguments`.

### DIFF
--- a/src/cmd_argument.rs
+++ b/src/cmd_argument.rs
@@ -1,5 +1,6 @@
 use crate::config::Config;
 use std::{
+    ffi::{OsStr, OsString},
     path::{Path, PathBuf},
     sync::Arc,
 };
@@ -21,12 +22,30 @@ where
     }
 }
 
+/// Arguments of type [`OsString`] are passed to the child process
+/// as arguments.
+impl CmdArgument for OsString {
+    #[doc(hidden)]
+    fn prepare_config(self, config: &mut Config) {
+        config.arguments.push(self);
+    }
+}
+
+/// Arguments of type [`OsStr`] are passed to the child process
+/// as arguments.
+impl CmdArgument for &OsStr {
+    #[doc(hidden)]
+    fn prepare_config(self, config: &mut Config) {
+        self.to_os_string().prepare_config(config);
+    }
+}
+
 /// Arguments of type [`&str`] are passed to the child process
 /// as arguments.
 impl CmdArgument for &str {
     #[doc(hidden)]
     fn prepare_config(self, config: &mut Config) {
-        config.arguments.push(self.into());
+        OsStr::new(self).prepare_config(config);
     }
 }
 
@@ -35,7 +54,7 @@ impl CmdArgument for &str {
 impl CmdArgument for String {
     #[doc(hidden)]
     fn prepare_config(self, config: &mut Config) {
-        config.arguments.push(self.into());
+        OsString::from(self).prepare_config(config);
     }
 }
 
@@ -252,7 +271,7 @@ where
 impl CmdArgument for PathBuf {
     #[doc(hidden)]
     fn prepare_config(self, config: &mut Config) {
-        config.arguments.push(self.into());
+        self.into_os_string().prepare_config(config);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/cmd_argument.rs
```diff
@@ -21,12 +22,30 @@ where
...
+/// Arguments of type [`OsString`] are passed to the child process
+/// as arguments.
+impl CmdArgument for OsString {
+    #[doc(hidden)]
+    fn prepare_config(self, config: &mut Config) {
+        config.arguments.push(self);
+    }
+}
+
+/// Arguments of type [`OsStr`] are passed to the child process
```
soenkehahn, line 34:
```suggestion
/// Arguments of type [`&OsStr`] are passed to the child process
```

src/cmd_argument.rs
```diff
@@ -21,12 +22,30 @@ where
     }
 }
 
+/// Arguments of type [`OsString`] are passed to the child process
+/// as arguments.
```
soenkehahn, line 26: Maybe a doctest would be good here? Maybe something with the example from the PR description, i.e. using environment variables...

src/cmd_argument.rs
```diff
@@ -21,12 +22,30 @@ where
...
+impl CmdArgument for OsString {
+    #[doc(hidden)]
+    fn prepare_config(self, config: &mut Config) {
+        config.arguments.push(self);
+    }
+}
+
+/// Arguments of type [`OsStr`] are passed to the child process
+/// as arguments.
```
soenkehahn, line 35: Maybe a doctest would be useful? I.e. with `Path::file_name`.

soenkehahn (commented):
Nice. Yeah, implementing everything in terms of `OsString` is nice.

Left some comments.

src/cmd_argument.rs
casey: I actually tried this, and for some reason it produces an error when running `cargo doc`:

```
warning: unresolved link to `&OsStr`
  --> src/cmd_argument.rs:40:24
   |
40 | /// Arguments of type [`&OsStr`] are passed to the child process
   |                        ^^^^^^^^ no item named `&OsStr` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`

warning: 1 warning emitted
```

It's a bit confusing, since `[&str]` works. Maybe `&str` is special, and otherwise it only works with identifiers and paths, i.e. with no leading `&`?

casey:
Added some doctests. I wasn't able to figure out the `&OsStr` doc link thing, kinda puzzling.

src/cmd_argument.rs
```diff
@@ -21,12 +22,42 @@ where
...
+///
+/// ```
+/// use cradle::*;
+///
+/// cmd_unit!("echo", std::env::current_dir().unwrap().file_name().unwrap());
```
soenkehahn, line 46: Same as above, I feel there should be a better example than just using `echo`. But I'm also having a hard time coming up with one... documentation is hard.

soenkehahn (approved):
Looks good. I've added some comments, but for the ones about better doctest examples, I think this is really hard and I wouldn't block merging this PR on that.

src/cmd_argument.rs
soenkehahn: Made this change myself.

casey:
Sweet, thanks for fixing and merging. How did you get `cargo doc` to stop complaining about `&OsStr`?

soenkehahn:
You can have links with arbitrary text in them by adding these link definitons: https://github.com/soenkehahn/cradle/pull/71/files#diff-314b0e8d3543865f2dac2e6e8ae1d3cc4abb8dadebf2a4576d8674c07c010ea4R49

PR #72, opened 2021-06-27
Remove unnecessary conversions

casey:
Removes some conversions which are unnecessary, now that cradle is more liberal in what it accepts as arguments.


soenkehahn:
Nice!

PR #73, opened 2021-06-27
Allow writing arbitrary byte sequences to stdin

casey:
This changes `Stdin` to accept anything that implements `AsRef<[u8]`>`, which includes `String`, `&str`, `Vec<u8>`, `&[u8]`, and probably other things.

It uses a single `Vec<u8>` and issues a single write call, so it should be very slightly faster.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -313,9 +313,6 @@ pub struct Stdin<T: AsRef<[u8]>>(pub T);
 
 /// Writes the given byte slice to the child's standard input.
 ///
-/// If `Stdin` is used multiple times, all given bytes slices will be written
-/// to the child's standard input in order.
-///
 /// ```
 /// use cradle::*;
 ///
@@ -325,6 +322,9 @@ pub struct Stdin<T: AsRef<[u8]>>(pub T);
 /// assert_eq!(output, "bar\nfoo\n");
 /// # }
 /// ```
+///
+/// If `Stdin` is used multiple times, all given bytes slices will be written
+/// to the child's standard input in order.
 impl<T> Input for Stdin<T>
 where
     T: AsRef<[u8]>,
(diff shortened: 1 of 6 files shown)


soenkehahn:
I'm a bit concerned about the negative effect on the generated documentation. From `impl CmdArgument for Into<String>` it's pretty clear that you can use this to pass in `String`s as `CmdArgument`. For `AsRef<[u8]>` maybe less so. Do you have a use-case where `Into<String>` wasn't good enough?

casey:
I think I would actually make the opposite argument. `std::process::Child::stdin` implements `std::io::Write`, to which arbitrary bytes can be written. So seeing that `cradle::Stdin` takes `Into<String>` makes me wonder if it's doing something extra or unexpected.

I don't have a specific use case for passing bytes to stdin, but I think there are common use cases that are bound to come up, like passing bytes to `shasum`, passing archives to `tar` on stdin, etc.

soenkehahn:
> I think I would actually make the opposite argument. `std::process::Child::stdin` implements `std::io::Write`, to which arbitrary bytes can be written. So seeing that `cradle::Stdin` takes `Into<String>` makes me wonder if it's doing something extra or unexpected.

Yeah, I buy that argument. Let's do it then. I'll try to update this branch.

src/input.rs
```diff
@@ -324,12 +322,15 @@ pub struct Stdin<T: Into<String>>(pub T);
 /// assert_eq!(output, "bar\nfoo\n");
 /// # }
 /// ```
+///
+/// If `Stdin` is used multiple times, all given bytes slices will be written
+/// to the child's standard input in order.
```
soenkehahn, line 327: I moved this sentence below the code example. I think that makes more sense. WDYT, @casey?

soenkehahn:
I updated the branch and added one question in a comment. Apart from that this LGTM for merging now.

src/input.rs
casey: Looks good to me!

casey:
Great, lgtm

PR #74, opened 2021-06-27
Parallelize tests and implement CmdArgument for TempDir

casey:
I did this because I thought the tests were slow because of `--test-threads=1`, but it turns out it's because [doctests are just inherently slow](https://github.com/rust-lang/cargo/issues/2944), so this doesn't actually improve the speed.

It does allow tests to be run with `cargo test` (at least if the test helper binaries are built), which is nice.

I also changed it to avoid creating tempfiles in the project root, and removed the corresponding entries in `.gitignore`.

To do this, I added a `CmdArgument` for `TempDir`, which I think is super useful, so I made it an optional feature.

You can enable the `tepdir` feature to enable the `CmdArgument` for `TempDir`, which enables the optional dependency on `tempfile`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,2 @@
 /target/
 Cargo.lock
-/foo
-/filename with spaces
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,4 +44,5 @@ path = "src/context_integration_tests.rs"
 required-features = ["test_executables"]
 
 [features]
+tempdir = ["tempfile"]
 test_executables = ["gag", "executable-path", "nix"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,10 @@ homepage = "https://github.com/soenkehahn/cradle"
 [dependencies]
 rustversion = "1.0.4"
 
+[dependencies.tempfile]
+version = "3.2.0"
+optional = true
+
 [dev-dependencies]
 executable-path = "1.0.0"
 pretty_assertions = "0.7.2"
--- a/justfile
+++ b/justfile
@@ -4,11 +4,10 @@ build:
   cargo build --all-targets --all-features
 
 test +pattern="": build
-  cargo test --all -- --test-threads=1 {{ pattern }}
-  rm -f 'filename with spaces' foo
+  cargo test --all -- {{ pattern }}
 
 test-lib-fast +pattern="":
-  cargo test --lib -- --test-threads=1 {{ pattern }}
+  cargo test --lib -- {{ pattern }}
 
 context-integration-tests: build
   cargo run --features "test_executables" --bin context_integration_tests
--- a/src/cmd_argument.rs
+++ b/src/cmd_argument.rs
@@ -305,7 +305,7 @@ where
 
 /// Argument of type [`TempDir`] configure the child to run
 /// with the current directory set to the tempdir.
-#[cfg(any(test, feature = "tempfile"))]
+#[cfg(any(test, feature = "tempdir"))]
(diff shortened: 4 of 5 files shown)


soenkehahn (commented):
Hmm, I think I'd prefer not making this change. For the doctests: I think that the main concern should be what would be the best documentation for users. And introducing a `tempdir` is not really useful and might even be confusing. And by default it doesn't even work for users, since the `tempdir` feature is not enabled. And for the unit tests: I think introducing a `tempdir` as an argument to `cmd!` is not really nice, because it moves the tests slightly further away from the real use-cases that are tested. And personally I have nothing against how `in_temporary_directory` works.

I do agree that it'd be nice to solve the problems this PR solves: Allowing `cargo test` to work out of the box and avoiding to write test files into the root directory of the repo. But I think there are different solutions for these problems, that don't come with the above-mentioned drawbacks. For example we could wrap the doctests in `in_temporary_directory` and comment the lines where that happens, so that they don't show up in the docs. And to allow `cargo test`, we could have a test wrapper (similar to `in_temporary_directory`) that acquires a global lock.

But yeah, you also wrote that the initial intention for this PR was to make the tests faster. I agree that that would be great! At the same time I also think that the two solvable problems from above don't have a really high priority. So I would propose to close this PR unmerged. If you feel that either of the two problems are really important, of course feel free to open issues or PRs for them.

What do you think?

casey:
> Hmm, I think I'd prefer not making this change. For the doctests: I think that the main concern should be what would be the best documentation for users. And introducing a `tempdir` is not really useful and might even be confusing. And by default it doesn't even work for users, since the `tempdir` feature is not enabled. And for the unit tests: I think introducing a `tempdir` as an argument to `cmd!` is not really nice, because it moves the tests slightly further away from the real use-cases that are tested. And personally I have nothing against how `in_temporary_directory` works.
> 
> I do agree that it'd be nice to solve the problems this PR solves: Allowing `cargo test` to work out of the box and avoiding to write test files into the root directory of the repo. But I think there are different solutions for these problems, that don't come with the above-mentioned drawbacks. For example we could wrap the doctests in `in_temporary_directory` and comment the lines where that happens, so that they don't show up in the docs. And to allow `cargo test`, we could have a test wrapper (similar to `in_temporary_directory`) that acquires a global lock.
> 
> But yeah, you also wrote that the initial intention for this PR was to make the tests faster. I agree that that would be great! At the same time I also think that the two solvable problems from above don't have a really high priority. So I would propose to close this PR unmerged. If you feel that either of the two problems are really important, of course feel free to open issues or PRs for them.
> 
> What do you think?

I definitely agree with a lot of your points. The doctests should be as good as possible, and definitely shouldn't rely on an optional feature. That could be done with, instead of having a `CmdArgument` implementation for `&Tempdir`, just using `CurrentDir`:

```rust
let tmpdir = Tempdir::new().unwrap();

cmd_unit!(%"touch foo", CurrentDir(tmpdir.path()));
```

I think that I find `in_temporary_directory` to just be "weird", and removing it would make the tests clearer to me, and closer to how tests in usually work. I kind of suspect that other rust developers might feel similarly, i.e. they expect tests to just run concurrently, and not modify the global current dir.

One non-aesthetic downside of `in_temporary_directory` is that if we made it take a lock, then it would be a bug to call the `std::env` functions without acquiring that lock, since a concurrent `in_temporary_directory` call might be munging them.

soenkehahn:
Closing in favor of solving these problems (parallelize tests, get rid of weird `in_temporary_directory`, not writing test files into the repo root) in subsequent PRs individually. (If we care enough.)

PR #75, opened 2021-06-27
Allow trailing commas in `cmd!` invocations

casey:
To make this work, I added a `$(,)?` to the end of the two base cases, and made them first, so they match in the case you get to the end and there's still a trailing comma.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1372,11 +1372,15 @@ mod tests {
         #[test]
         fn trailing_comma_is_accepted_after_normal_argument() {
             cmd_unit!("echo", "foo",);
+            let () = cmd!("echo", "foo",);
+            let _result: Result<(), Error> = cmd_result!("echo", "foo",);
         }
 
         #[test]
         fn trailing_comma_is_accepted_after_split_argument() {
             cmd_unit!("echo", %"foo",);
+            let () = cmd!("echo", %"foo",);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1365,4 +1365,18 @@ mod tests {
...
+
+        #[test]
+        fn trailing_comma_is_accepted_after_normal_argument() {
+            cmd_unit!("echo", "foo",);
+        }
+
+        #[test]
+        fn trailing_comma_is_accepted_after_split_argument() {
+            cmd_unit!("echo", %"foo",);
+        }
```
soenkehahn, line 1380: Maybe it'd be good to also have one test for `cmd!` and one for `cmd_result!`. So that all unhidden macros are covered.

soenkehahn (commented):
Nice. Added one comment, apart from that LGTM.

casey:
Sounds good, added tests for `cmd` and `cmd_result`.

soenkehahn:
Nice, thanks! The CI is failing with a clippy warning on older rustc versions. I think it can be fixed by doing `let StdoutUntrimmed(_) = cmd!("echo", %"foo",);`.

(I can do that, if you want.)

casey:
Ahh, thanks. Should be fixed now.

soenkehahn:
Awesome.

soenkehahn:
Published as https://crates.io/crates/cradle/0.0.6.

PR #83, opened 2021-07-01
Remove unnecessary wrapping in Vec in tests

soenkehahn:
This used to be necessary in tests to avoid splitting on whitespace.
But `cradle` doesn't split strings by whitespace by default anymore.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -414,7 +414,7 @@ mod tests {
             #[test]
             #[should_panic(expected = "exited with exit code: 42")]
             fn other_exit_codes() {
-                cmd_unit!(executable_path("cradle_test_helper"), vec!["exit code 42"]);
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -605,13 +605,6 @@ mod tests {
         assert_eq!(stdout, "foo");
     }
 
-    #[test]
-    fn allows_to_pass_in_arguments_as_a_vec_of_ref_str() {
-        let args: Vec<&str> = vec!["foo"];
-        let StdoutTrimmed(stdout) = cmd!("echo", args);
-        assert_eq!(stdout, "foo");
-    }
-
```
soenkehahn, line 614: This test somehow got duplicated with the one in `sequences`.

PR #88, opened 2021-07-07
Implement support for tuples as Inputs

soenkehahn:


src/input.rs
```diff
@@ -43,6 +43,51 @@ use std::{
 /// [`CurrentDir`]: trait.Input.html#impl-Input-1
 /// [`StdIn`]: trait.Input.html#impl-Input-2
 /// [`LogCommand`]: trait.Input.html#impl-Input
+///
+/// ## Tuples
+///
+/// `cradle` also provides implementations for [`Input`] for tuples of different lengths.
```
casey, line 49:
```suggestion
/// `cradle` also implements [`Input`] for tuples of types that themselves implement [`Input`].
```

src/input.rs
```diff
@@ -43,6 +43,51 @@ use std::{
...
+/// So instead of passing in different arguments into [`cmd!`] seperately,
+/// you could put them into a tuple:
```
casey, line 51:
```suggestion
/// Instead of passing multiple arguments to [`cmd!`], they can be passed in a single tuple:
```

src/input.rs
```diff
@@ -43,6 +43,51 @@ use std::{
...
+///
+/// ```
+/// use cradle::*;
+///
+/// let args = ("echo", "foo");
+/// let StdoutTrimmed(output) = cmd!(args);
+/// assert_eq!(output, "foo");
+/// ```
+///
+/// This is useful to allow to bundle up arguments for convenience:
```
casey, line 61:
```suggestion
/// This can be used to group arguments:
```

src/input.rs
```diff
@@ -43,6 +43,51 @@ use std::{
...
+///
+/// ```
+/// use cradle::*;
+///
+/// let to_hex_command = ("xxd", "-ps", "-u", LogCommand);
+/// let StdoutTrimmed(output) = cmd!(to_hex_command, Stdin(&[14, 15, 16]));
+/// assert_eq!(output, "0E0F10");
+/// ```
+///
+/// Also, tuples allow to write flexible wrappers around [`cmd!`] without using macros:
```
casey, line 71:
```suggestion
/// Also, tuples make it possible to write wrappers around [`cmd!`] without requiring the use of macros:
```

src/input.rs
```diff
@@ -43,6 +43,51 @@ use std::{
...
+///
+/// ```
+/// use cradle::*;
+///
+/// fn to_hex<I: Input>(input: I) -> String {
+///   let StdoutTrimmed(hex) = cmd!(%"xxd -ps -u", input);
+///   hex
+/// }
+///
+/// // Works for slices:
```
casey, line 81:
```suggestion
/// // It works for slices:
```

src/input.rs
```diff
@@ -43,6 +43,51 @@ use std::{
...
+/// let hex = to_hex(Stdin(&[14, 15, 16]));
+/// assert_eq!(hex, "0E0F10");
+/// // Also works for vectors:
```
casey, line 84:
```suggestion
///
/// // Vectors:
```

src/input.rs
```diff
@@ -43,6 +43,51 @@ use std::{
...
+/// let hex = to_hex(Stdin(vec![14, 15, 16]));
+/// assert_eq!(hex, "0E0F10");
+/// // Also works for multiple arguments using tuples:
```
casey, line 87:
```suggestion
///
/// // And multiple arguments using tuples:
```

src/lib.rs
```diff
@@ -984,6 +984,35 @@ mod tests {
...
+
+        #[test]
+        fn nested_tuples() {
+            let StdoutTrimmed(output) = cmd!(("echo", ("foo", "bar")));
+            assert_eq!(output, "foo bar");
+        }
+
+        #[test]
+        fn unit_input() {
+            let StdoutTrimmed(output) = cmd!(("echo", ()));
```
casey, line 1011: It's an interesting consequence, maybe useful although I can't think of how immediately, that `()` as `Input` is a noop.

casey (commented):
Comment comments but otherwise looks good to me.

src/lib.rs
soenkehahn: You could imagine a wrapper function like this:
``` rust
fn run_docker<I: Input>(additional_args: I) {
  cmd_unit!(%"docker run", additional_args, "ubuntu:latest", "bash");
}
```
The function always performs the same `docker` command, but allows to tweak it by passing in additional arguments. And if callers want to not tweak anything, they can pass in `()`.

casey: Good example!

PR #89, opened 2021-07-08
Add documentation for Input

soenkehahn:


src/input.rs
```diff
@@ -5,7 +5,44 @@ use std::{
     sync::Arc,
 };
 
-/// All types that are possible arguments to [`cmd!`] have to implement this trait.
+/// All types that are possible arguments to [`cmd!`], [`cmd_unit!`] or
+/// [`cmd_result!`] have to implement this trait.
```
casey, line 9:
```suggestion
/// All  types that are possible arguments to [`cmd!`], [`cmd_unit!`] or
/// [`cmd_result!`] must implement this trait.
```

src/input.rs
```diff
@@ -5,7 +5,44 @@ use std::{
...
+/// This makes `cradle` very flexible.
+/// For example passing in an executable can be done as a String,
+/// and passing in a couple of arguments can be done in a [`Vec`]:
```
casey, line 12:
```suggestion
/// and passing in a variable of arguments can be done in a [`Vec`]:
```

Couple is fine, but for slightly formal technical documentation, "a variable number of" might be better.

src/input.rs
```diff
@@ -5,7 +5,44 @@ use std::{
...
+/// use cradle::*;
+///
+/// let executable = "echo";
+/// let arguments = vec!["foo", "bar"];
+/// let StdoutUntrimmed(output) = cmd!(executable, arguments);
+/// assert_eq!(output, "foo bar\n");
+/// ```
+///
+/// For more documentation on all possible input types,
+/// see to the documentation for the individual impls for [`Input`].
```
casey, line 24:
```suggestion
/// see the documentation for the individual impls of [`Input`].
```

src/input.rs
```diff
@@ -5,7 +5,44 @@ use std::{
...
+/// Here's a non-exhaustive list of the most commonly used types to get you started:
```
casey, line 25: Listing them out like this is a great idea for discoverability.

casey (commented):
LGTM!

PR #91, opened 2021-07-10
Tweak documentation for Output

soenkehahn:
This PR tweaks the documentation for `Output`. Also -- since it's now easier to wade through the impls of `Output` -- it adds impls for longer tuples.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -1,23 +1,54 @@
 use crate::{Config, Error, RunResult};
 use std::{process::ExitStatus, sync::Arc};
 
-/// All possible return types of [`cmd!`] have to implement this trait.
-/// For documentation about what these return types do, see the
-/// individual implementations below.
-///
-/// Except for tuples: All [`Output`] implementations for tuples serve
-/// the same purpose: combining multiple types that implement [`Output`]
-/// to retrieve more information from a child process. The following code
-/// for example retrieves what's written to `stdout` **and** the
-/// [`ExitStatus`]:
+/// All possible return types of [`cmd!`], [`cmd_unit!`] or
+/// [`cmd_result!`] must implement this trait.
+/// This return-type polymorphism makes cradle very flexible.
+/// For example if you want to capture what a command writes
+/// to its `stdout` you can do that using [`StdoutUntrimmed`]:
 ///
 /// ```
 /// use cradle::*;
 ///
-/// let (StdoutUntrimmed(stdout), Exit(status)) = cmd!(%"echo foo");
-/// assert_eq!(stdout, "foo\n");
-/// assert!(status.success());
+/// let StdoutUntrimmed(output) = cmd!(%"echo foo");
+/// assert_eq!(output, "foo\n");
 /// ```
+///
+/// But if instead you want to capture the commands exit code,
+/// you can use [`Exit`]:
+///
+/// ```
+/// use cradle::*;
+///
+/// let Exit(status) = cmd!("false");
+/// assert_eq!(status.code(), Some(1));
+/// ```
+///
+/// For documentation on what all these possible return types do,
+/// see the documentation for the individual impls of [`Output`].
+/// Here's a non-exhaustive list of the more commonly used return-types to get you started:
+///
+/// - `()`: In case you don't want to capture anything. See also [`cmd_unit`].
+/// - [`StdoutUntrimmed`], [`StdoutTrimmed`] and [`Stderr`]: To capture `stdout` and `stderr`.
+/// - [`Exit`]: To capture the commands exit code.
+///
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -1,23 +1,54 @@
...
 use std::{process::ExitStatus, sync::Arc};
 
-/// All possible return types of [`cmd!`] have to implement this trait.
-/// For documentation about what these return types do, see the
-/// individual implementations below.
+/// All possible return types of [`cmd!`], [`cmd_unit!`] or
+/// [`cmd_result!`] must implement this trait.
+/// This return-type polymorphism makes cradle very flexible.
+/// For example if you want to capture what a command writes
+/// to its `stdout` you can do that using [`StdoutUntrimmed`]:
```
casey, line 8:
```suggestion
/// For example, if you want to capture what a command writes
/// to `stdout` you can do that using [`StdoutUntrimmed`]:
```

src/output.rs
```diff
@@ -1,23 +1,54 @@
...
-/// for example retrieves what's written to `stdout` **and** the
-/// [`ExitStatus`]:
+/// ```
+/// use cradle::*;
+///
+/// let StdoutUntrimmed(output) = cmd!(%"echo foo");
+/// assert_eq!(output, "foo\n");
+/// ```
+///
+/// But if instead you want to capture the commands [`ExitStatus`],
```
casey, line 17:
```suggestion
/// But if instead you want to capture the command's [`ExitStatus`],
```

src/output.rs
```diff
@@ -1,23 +1,54 @@
...
+/// let Exit(status) = cmd!("false");
+/// assert_eq!(status.code(), Some(1));
+/// ```
+///
+/// For documentation on what all the possible return types do,
+/// see the documentation for the individual impls of [`Output`].
+/// Here's a non-exhaustive list of the more commonly used return types to get you started:
+///
+/// - `()`: In case you don't want to capture anything. See also [`cmd_unit`].
+/// - [`StdoutUntrimmed`], [`StdoutTrimmed`] and [`Stderr`]: To capture `stdout` and `stderr`.
```
casey, line 32:
```suggestion
/// - [`StdoutUntrimmed`], [`StdoutTrimmed`] and [`Stderr`]: To capture `stdout`, `stdout` trimmed of whitespace, and `stderr`.
```

src/output.rs
```diff
@@ -1,23 +1,54 @@
...
+/// - [`Exit`]: To capture the commands [`ExitStatus`].
```
casey, line 33:
```suggestion
/// - [`Exit`]: To capture the command's [`ExitStatus`].
```

src/output.rs
```diff
@@ -1,23 +1,54 @@
...
+///
+/// Alse, [`Output`] is implemented for a number of tuples of different lengths.
```
casey, line 35:
```suggestion
/// Alse, [`Output`] is implemented for tuples.
```

src/output.rs
```diff
@@ -1,23 +1,54 @@
...
+/// You can use this to combine other return types that implement [`Output`].
```
casey, line 36:
```suggestion
/// You can use this to combine multiple return types that implement [`Output`].
```

src/output.rs
```diff
@@ -1,23 +1,54 @@
...
+/// The following code for example retrieves the [`ExitStatus`]
+/// **and** what's written to `stdout`:
```
casey, line 38:
```suggestion
/// The following code for example retrieves the command's [`ExitStatus`]
/// **and** what it writes to`stdout`:
```

src/output.rs
```diff
@@ -1,23 +1,54 @@
...
+///
+/// ```
+/// use cradle::*;
+///
+/// let (Exit(status), StdoutUntrimmed(stdout)) = cmd!(%"echo foo");
 /// assert!(status.success());
+/// assert_eq!(stdout, "foo\n");
 /// ```
+///
+/// [`StdoutUntrimmed`]: trait.Output.html#impl-Output-3
```
casey, line 48: Will these links remain stable as long as the order doesn't change? Is there any way to link to impls by name?

casey:
Bunch of comment comments, but otherwise looks good.

src/output.rs
soenkehahn: I wondered the same thing. No idea whether they will stay stable. And I didn't find a better way to link to the trait implementation for specific types. I don't think it exists, looking at [this](https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html#valid-links).

My plan was to see whether these stay stable when adding another type to either `Input` or `Output`. And look for a better solution if they don't.

PR #95, opened 2021-07-12
Put primary macros into a `prelude` submodule

soenkehahn:
This is an attempt to put the primary interface -- the items that most people will use and need to import -- into a `prelude` submodule. So that users can do `use cradle::prelude::*;`. This has some disadvantages, so I don't think this will ever get merged. The dealbreaker seems to be this:

Even if macros are declared in a submodule, rust puts them into the top module of a crate. Here I'm using this re-exporting hack to work around that, but that seems like a terrible idea. Also, I haven't checked what it does to generated documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 Here's an example:
 
 ``` rust
-use cradle::*;
+use cradle::prelude::*;
 
 fn main() {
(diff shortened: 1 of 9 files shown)


soenkehahn:
I couldn't find a way to make this actually work well in the generated documentation.

PR #97, opened 2021-07-15
Rename `cmd_unit` to `cmd_void`

soenkehahn:
I think `cmd_void` is good, since it mirrors what `void` means in e.g. C or Java. Also, even if people don't see that connection, you can think of this as sending any possible `Output`s into the void.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -5,7 +5,7 @@ use std::{
     sync::Arc,
 };
 
-/// All types that are possible arguments to [`cmd!`], [`cmd_unit!`] or
+/// All types that are possible arguments to [`cmd!`], [`cmd_void!`] or
 /// [`cmd_result!`] must implement this trait.
 /// This makes `cradle` very flexible.
 /// For example you can pass in an executable as a String,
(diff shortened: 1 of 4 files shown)


casey:
I'm not sure about this!

I think `cmd_void` would be better in Java or C, and it might be more immediately familiar to Rust programmers who know those languages. However, I think `cmd_unit` is better overall, even if it does have problems. Not all programmers are familiar with C or Java, so for those programmers, it's not a ton of help. Also, unit / `()` is a Rust language concept, so `cmd_unit` is better for programmers that know about unit, and for those that don't, it's a good learning experience (less important) and when they do learn it, it'll fit better with other things in the language (more important).

soenkehahn:
Yeah, I just found this: https://doc.rust-lang.org/std/primitive.unit.html. If we link to that somewhere, then `cmd_unit!` is actually a good name.

casey:
Sounds good, I think linking to that doc is a great idea.

soenkehahn:
Came up with a PR here: #100.

PR #98, opened 2021-07-15
Quote emtpy args when using `LogCommand`

soenkehahn:


casey:
I think it might be better to quote all arguments for consistency, but this is definitely a good change, so LGTM.

PR #100, opened 2021-07-16
Link to unit type in docs of `cmd_unit`

soenkehahn:


src/lib.rs
```diff
@@ -91,8 +91,9 @@
 //! let () = cmd!(%"touch foo");
 //! ```
 //!
-//! Since that's a very common case, `cradle` provides the [`cmd_unit!`]
-//! shortcut, that behaves exactly like [`cmd!`], but pins the return
+//! Since that's a very common case, `cradle` provides the [`cmd_unit!`] shortcut.
+//! It's named after [the unit type `()`](https://doc.rust-lang.org/std/primitive.unit.html).
+//! It behaves exactly like [`cmd!`], but pins the return
 //! type down to `()`:
```
casey, line 97:
```suggestion
//! Since that's a very common case, `cradle` provides the [`cmd_unit!`] shortcut.
//! It's named after [the unit type `()`](https://doc.rust-lang.org/std/primitive.unit.html).
//! It behaves exactly like [`cmd!`] but always returns `()`.
```

casey:
Minor suggestion, otherwise LGTM

PR #104, opened 2021-07-17
Add categories to Cargo.toml

soenkehahn:
I think these are the most fitting of the available [categories](https://crates.io/categories). Not sure whether it's better to add them or leave them off though.


casey:
I think these are good. They're not perfect, but that's fine.

PR #110, opened 2021-07-20
Allow to set environment variables

soenkehahn:


src/input.rs
```diff
@@ -25,23 +25,25 @@ use std::{
...
 /// [`PathBuf`]: trait.Input.html#impl-Input-for-PathBuf
 /// [`&Path`]: trait.Input.html#impl-Input-for-%26Path
 /// [`vectors`]: trait.Input.html#impl-Input-for-Vec<T>
 /// [`slices`]: trait.Input.html#impl-Input-for-%26[T]
 /// [`arrays`]: trait.Input.html#impl-Input-for-[T%3B%20N]
-/// [`CurrentDir`]: trait.Input.html#impl-Input-1
-/// [`StdIn`]: trait.Input.html#impl-Input-2
+/// [`CurrentDir`]: trait.Input.html#impl-Input-2
+/// [`SetVar`]: trait.Input.html#impl-Input-1
+/// [`StdIn`]: trait.Input.html#impl-Input-3
```
soenkehahn, line 46: Unfortunately these links are not stable when adding impls. I don't want to fix this in this PR, but I opened this issue instead: #111.

src/lib.rs
```diff
@@ -288,8 +290,11 @@ where
             .map_err(|error| Error::command_io_error(&config, error))?;
     }
     let mut command = Command::new(&executable);
+    command.args(arguments);
+    for (key, value) in &config.environment_additions {
+        command.env(key, value);
+    }
     command
-        .args(arguments)
         .stdin(Stdio::piped())
```
casey, line 298: I haven't tested this, but I think it should work:

```suggestion
        .args(arguments)
        .envs(&config.environment_additions)
        .stdin(Stdio::piped())
```

casey (commented):
A few minor comments, but otherwise looks great. I like `Var` better than `SetVar`, but `SetVar` is okay too. I think `Env` is also okay, to mirror `Command::env`, although I think `Env` is a little less clear.

casey:
Actually, thinking about it, I think using the same names as the `Command` methods is probably a good idea. There a few methods related to environment variables: `Command::{env, envs, env_clear, env_remove}`. Although I don't think that they have the best names, just reusing the names for `cradle` will probably be the easiest to learn, and the easiest to relate back to `Command` methods.

soenkehahn:
Yes, I agree with respect to `Command::...` names.

src/lib.rs
soenkehahn: I couldn't make this work. `Command::envs` expects an iterator over tuples, not references to tuples. So we could clone the tuple, of course. But that seems worse, since with the for loop above, we don't have to clone anything.

src/input.rs
```diff
@@ -460,3 +462,36 @@ where
         Arc::make_mut(&mut config.stdin).extend_from_slice(self.0.as_ref());
     }
 }
+
+/// See the [`Input`] implementation for [`SetVar`] below.
+pub struct SetVar<Key, Value>(pub Key, pub Value)
```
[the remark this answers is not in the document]
soenkehahn: Going with `Env`, which matches the `Process` interface.

PR #114, opened 2021-07-21
Implement `Output` for `bool`

casey:


src/output.rs
```diff
@@ -205,6 +205,45 @@ impl Output for Status {
...
 }
 
+/// Using [`bool`] as the return type for [`cmd!`] will return `true` if
+/// the command returned successfully, and `false` otherwise:
+///
+/// ```
+/// use cradle::*;
+///
+/// if !cmd!(%"echo foo") {
+///     panic!("echo failed!");
```
soenkehahn, line 215:
```suggestion
///     panic!("echo should not fail!");
```

src/lib.rs
```diff
@@ -992,6 +992,26 @@ mod tests {
...
+        #[test]
+        fn failure_exit_status_is_false() {
+            assert!(!cmd!("false"));
+        }
+
+        #[test]
+        #[should_panic]
+        fn io_error_panics() {
+            assert!(cmd!("/"));
+        }
```
soenkehahn, line 1012: Nice tests. This might not _always_ be the behavior people want or expect. But it's definitely good to document it in a test.

src/output.rs
```diff
@@ -205,6 +205,45 @@ impl Output for Status {
...
+/// }
```
soenkehahn, line 216: I wonder if we can come up with a better example. Some command that actually can return both `true` and `false` in different circumstances, maybe?

soenkehahn (commented):
LGTM, apart from comments. (I might put some commits into the branch.)

src/output.rs
```diff
@@ -251,3 +251,42 @@ impl Output for Status {
...
+/// or [`std::result::Result::Err`]:
+///
+/// ```
+/// use cradle::*;
+///
+/// let success: bool = cmd!("false");
+/// assert!(!success);
+/// let result: Result<bool, cradle::Error> = cmd_result!("false");
+/// assert!(result.is_ok());
+/// assert!(!result.unwrap());
```
soenkehahn, line 276:
```suggestion
/// assert_eq!(result, Some(false));
```

I _think_ this is slightly better. TIOLI.

src/output.rs
```diff
@@ -251,3 +251,42 @@ impl Output for Status {
...
+
+/// Using [`bool`] as the return type for [`cmd!`] will return `true` if
+/// the command returned successfully, and `false` otherwise:
+///
+/// ```
+/// use cradle::*;
+///
+/// if !cmd!(%"echo foo") {
+///     panic!("echo should not fail!");
+/// }
```
soenkehahn, line 263:
```suggestion
/// if !cmd!(%"which cargo") {
///     panic!("`cargo` is not installed");
/// }
```

Maybe this is better?

casey: Agreed, I was actually thinking that `assert!`-ing a bool isn't super clear.

src/output.rs
casey: Good call, I used `which cargo`.

casey:
Addressed review comments and rebased onto latest master.

casey:
I fixed the tests, if it passes it should be good to go. (I used `assert!(matches!(…))` because `Error` doesn't and can't support `==`.)

casey:
100% guaranteed to pass now.

casey:
🎉

PR #124, opened 2021-07-27
Output better source locations on panics

soenkehahn:


casey:
Nice test! Looks good to me.

PR #125, opened 2021-07-27
Put lock in in_temporary_directory to allow parallel test execution

soenkehahn:


src/lib.rs
```diff
@@ -364,13 +364,18 @@ mod tests {
...
         path::PathBuf,
+        sync::Mutex,
     };
     use tempfile::TempDir;
 
     fn in_temporary_directory<F>(f: F)
     where
         F: FnOnce() + std::panic::UnwindSafe,
     {
+        lazy_static::lazy_static! {
```
casey, line 375: I might `use lazy_static::lazy_static`, TIOLÍ.

casey (commented):
Looks good to me!

src/lib.rs
soenkehahn: Sounds good.

Issue #117, opened 2021-07-22
Make `cargo test` work


casey:
Currently, `cargo test` doesn't work out of the box, and extra flags are needed. The simplest way to make this work is probably to just add a mutex to `in_temp_dir`.

soenkehahn:
Unfortunately `cargo test` does not work out of the box, even after merging #125. The reason is that the tests rely on test helper executables that have to be built before running the test suite with e.g. `cargo build --all-targets --all-features`. But I think this issue was meant to get rid of the need for `--test-threads=1`. And #125 does that.

casey:
Ahh, that's too bad. One option to fix that would be something like this:

```rust
fn test_helper() -> PathBuf {
  let path = executable_path("cradle_test_helper");
  if !path.exists() {
    cmd_unit!(%"cargo build --bin cradle_test_helper --features test_executables");
  }
  path
}
```

And have tests get the path to the helper by running `test_helper()`.

soenkehahn:
Sadly, I think that only works when you don't change the `cradle_test_helper`. As soon as you start modifying the behavior of that file, this `.exists()` check will see that the old version is still in place and not cause a recompilation.

We could look at the time stamps, and recompile when the modification time of the source files is newer than the executable. But then we're pretty much in the business of writing a small build system, and I'm a bit hesitant to do that.

Here's two ideas:

1. We make sure (maybe through a `test_helper()` function using a `Once` value) that we once call `cargo build --bin cradle_test_helper --features test_executables`. And then rely on `cargo`'s ability to execute that very quickly if none of the source files have changed.
2. We use `rust-script` to execute the test-helper. Since hashbangs don't work on windows, it'd have to be something like `cmd_unit!("rust-script", test_helper(), ...)`. And then we would rely on `rust-script`'s ability to recompile quickly if nothing has changed. One thing that I haven't been able to figure out is how to make `rust-script` a dev-dependency, so that its executables are in the path while the tests are running. Is that possible?

casey:
I think 1 sounds the best. I'm guessing that it's very fast, and probably what `rust-script` does itself. Making dev-dependency binaries available isn't yet supported, see [this issue](https://github.com/rust-lang/cargo/issues/2267). (Opened by me in 2016!)

PR #129, opened 2021-07-27
Build test executables from within the test suite

soenkehahn:
This gets rid of the requirement to run `cargo build --bin cradle_test_helper --features test_executables` and `cargo build --bin cradle_panic --features test_executables` manually before running the test suite. Those commands are now triggered once per test-suite from within the test suite itself. More discussion here: #117


casey:
Nice, this looks good to me.

PR #133, opened 2021-07-29
Add a prelude module

soenkehahn:
- Add cradle::prelude module
- Add module documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 Here's an example:
 
 ``` rust
-use cradle::*;
+use cradle::prelude::*;
 
 fn main() {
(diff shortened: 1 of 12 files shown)


casey:
This looks good!

Separate issue, but should the `config` and `context` modules have `#[doc(hidden)]`? I was looking at the docs and they show up in the crate root, even though they're not supposed to be used.

soenkehahn:
> Separate issue, but should the `config` and `context` modules have `#[doc(hidden)]`? I was looking at the docs and they show up in the crate root, even though they're not supposed to be used.

Good point. I think yes. And I think this belongs in this PR, since before this, they didn't show up.

PR #136, opened 2021-07-29
Implement dedicated error variant when executable file cannot be found

soenkehahn:


casey:
This is an edge case, but sometimes the error message, `cannot find executable`, will be misleading. I poked around a bit, and not-found errors can occur because, for example, a dynamically linked library could not be found. It might be better to keep the error generic, something like:

```
File not found error when executing `BINARY`: IO-ERROR
```

Here's an example of a someone being befuddled about a file not found when the executable does exist:

https://unix.stackexchange.com/questions/378301/linux-executable-fails-with-file-not-found-even-though-the-file-is-there-and-i

Aside from that this looks good!

soenkehahn:
Good point. I'll change it.

PR #139, opened 2021-07-30
Add .run() and friends to Input

### LINKED ISSUE #96: Add `run` method to `Input`
We discussed this a little bit, although I don't think we came to a definitive conclusion.

It might be nice to add a `run` method to `Input`, to allow doing things like this:

```rust
("ls", std::env::current_dir().unwrap()).run();
```

soenkehahn:
This PR adds `.run()`, `.run_unit()` and `.run_result()` to the Input trait. This fixes #96. That issue doesn't mention `run_unit` and `run_result`, but I think those are necessary.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -97,6 +97,13 @@ pub trait Input {
     {
         crate::cmd!(self)
     }
+
+    fn run_unit(self)
+    where
+        Self: Sized,
+    {
+        crate::cmd_unit!(self);
+    }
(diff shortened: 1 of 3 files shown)


casey:
This looks good, I like this change a lot. We talked about names, and trying to harmonize the names between the macros and the methods, but it seems like a good idea to defer that and do it in a dedicated renaming PR.

PR #140, opened 2021-07-30
Replace test_executables/helper with python snippets

soenkehahn:


soenkehahn:
Working on this in earnest in #210. I'll keep this open for reference until that lands.

PR #142, opened 2021-08-03
Disable tests that write invalid utf-8 to stdout or stderr on windows

soenkehahn:
According to https://doc.rust-lang.org/std/io/struct.Stderr.html#note-windows-portability-consideration these tests shouldn't work on windows. They also didn't work in my tests. They _did_ work on github ci, and I don't know why. In any case I'm disabling these tests on windows.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -552,6 +552,7 @@ mod tests {
             }
 
             #[test]
+            #[cfg(not(windows))]
             fn invalid_utf8_to_stdout_is_allowed_when_not_captured() {
                 cmd_unit!(test_helper(), "invalid utf-8 stdout");
             }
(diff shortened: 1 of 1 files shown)


casey:
Looks good to me.

Huh, interesting. How did you find out that writing non-UTF-8 bytes to the windows console was an error?

soenkehahn:
Those tests didn't pass on my windows machine. Only on github CI, for some reason.

casey:
> Those tests didn't pass on my windows machine. Only on github CI, for some reason.

That makes sense, I'm guessing that on CI stderr is hooked up to a pipe from some collector process, and on your laptop it's hooked up to a console.

Issue #26, opened 2021-05-08
Don't collect `stdout` and `stderr` into memory when not needed


soenkehahn:
Currently, both `stdout` and `stderr` of the child process are being collected into memory by `cmd!`. This is necessary, for example for this to work:

``` rust
let (stdout, Stderr(stderr)): (String, Stderr) = cmd!(...);
```

But when those outputs are not being returned (for example with `cmd_unit`), the library should avoid collecting them, to save memory.

(In a lot of cases, the additional memory consumption of the current behavior is probably negligible. But you could imagine someone using `cmd!` to stream large amounts of data to `stdout`. In those cases, this optimization is necessary.)

soenkehahn:
Half-solved for `stdout` in #143.

PR #143, opened 2021-08-04
Don't read stdout into memory when not needed

soenkehahn:
`cradle` needs to read what's written to `stdout` into memory when the user wants to capture `stdout`, e.g. with `StdoutUntrimmed`. However currently it also reads all of `stdout` into memory when not capturing `stdout`, which is unnecessary. It's also conceivable (when there's lots of output) that this may cause problems. This PR avoids collecting `stdout` into memory when it's not captured.

The same could be done for `stderr`. Not sure whether that's as important as doing it for `stdout`. In any case it should be done in a separate PR.

This PR is half a fix for #26.


src/collected_output.rs
```diff
@@ -7,7 +7,7 @@ use std::{
 
 pub(crate) struct Waiter {
     stdin: JoinHandle<io::Result<()>>,
-    stdout: JoinHandle<io::Result<Vec<u8>>>,
+    stdout: JoinHandle<io::Result<Option<Vec<u8>>>>,
```
soenkehahn, line 10: We have an invariant now: this will only be `None` when `stdout` is *not* captured and `Some(...)` when `stdout` is captured. We're introducing a new `Error::CradleBug` in case this invariant is ever unexpectedly broken.

src/collected_output.rs
```diff
@@ -30,8 +30,8 @@ impl Waiter {
         });
         let mut context_clone = context.clone();
         let relay_stdout = config.relay_stdout;
-        let stdout_join_handle = thread::spawn(move || -> io::Result<Vec<u8>> {
-            let mut collected_stdout = Vec::new();
+        let stdout_join_handle = thread::spawn(move || -> io::Result<Option<Vec<u8>>> {
+            let mut collected_stdout = if relay_stdout { None } else { Some(Vec::new()) };
```
soenkehahn, line 34: Currently `relay_stdout` decides whether `stdout` is captured or not. So `relay_stdout == false` means it's captured and `relay_stdout == true` means, it's not. This might change in the future and there may be ways to either both capture and relay `stdout` or to both *not* capture and *not* relay it. But that's not implemented yet, so I think we should only have one boolean in `Config`. I do wonder whether it's more intuitive to have the boolean field be `capture_stdout` instead.

memory-test/src/bin/cradle_user.rs
```diff
@@ -0,0 +1,8 @@
+use cradle::prelude::*;
+
+fn main() {
+    let mut args = std::env::args();
+    let bytes: usize = args.nth(1).unwrap().parse().unwrap();
+    eprintln!("consuming {} kB", bytes / 2_usize.pow(10));
```
casey, line 6:
```suggestion
    eprintln!("consuming {} KiB", bytes / 2_usize.pow(10));
```

memory-test/src/bin/produce_bytes.rs
```diff
@@ -0,0 +1,21 @@
+use anyhow::Result;
+use std::io::{stdout, Write};
+
+fn main() -> Result<()> {
+    let mut args = std::env::args();
+    let mut bytes: usize = args.nth(1).unwrap().parse()?;
+    eprintln!("producing {} kB", bytes / 2_usize.pow(10));
```
casey, line 7:
```suggestion
    eprintln!("producing {} KiB", bytes / 2_usize.pow(10));
```

memory-test/src/bin/cradle_user.rs
casey, line 6:
```suggestion
    eprintln!("consuming {} KiB", bytes / 2_usize.pow(10));
```

memory-test/src/bin/produce_bytes.rs
casey, line 7:
```suggestion
    eprintln!("producing {} KiB", bytes / 2_usize.pow(10));
```

memory-test/src/bin/run_test.rs
```diff
@@ -0,0 +1,56 @@
...
+        .arg(bytes.to_string())
+        .stdout(Stdio::null())
+        .output()?;
+    let stderr = String::from_utf8(output.stderr)?;
+    eprintln!("{}", stderr);
+    if !output.status.success() {
+        panic!("running 'cradle_user' failed");
+    }
+    let memory_size_prefix = "Maximum resident set size (kbytes): ";
+    let kilo_bytes: usize = strip_prefix(
```
casey, line 36:
```suggestion
    let kibibytes: usize = strip_prefix(
```

src/collected_output.rs
```diff
@@ -30,8 +30,8 @@ impl Waiter {
...
             let buffer = &mut [0; 256];
```
casey, line 35: This buffer seems a little small, although that's not based on anything objective, just a hunch. Maybe `4096` instead?

src/error.rs
```diff
@@ -26,6 +26,10 @@ pub enum Error {
         full_command: String,
         source: Arc<FromUtf8Error>,
     },
+    CradleBug {
```
casey, line 29: I slightly prefer `Internal` as the name for this error, since internal errors come up in other contexts, like http 500s, although I don't hate `CradleBug`. In that case I might call it just `Bug`, since it's already a `cradle::Error`.

```suggestion
    Internal {
```

src/error.rs
```diff
@@ -70,24 +82,35 @@ impl Display for Error {
...
+            InvalidUtf8ToStdout { full_command, .. } => {
                 write!(f, "{}:\n  invalid utf-8 written to stdout", full_command)
             }
-            Error::InvalidUtf8ToStderr { full_command, .. } => {
+            InvalidUtf8ToStderr { full_command, .. } => {
                 write!(f, "{}:\n  invalid utf-8 written to stderr", full_command)
             }
+            CradleBug { .. } => {
+                let snippets = vec![
+                  "Congratulations, you've found a bug in cradle! :/",
```
casey, line 93: Love it

src/error.rs
```diff
@@ -70,24 +82,35 @@ impl Display for Error {
...
+                  "Please, consider reporting a bug on https://github.com/soenkehahn/cradle/issues,",
```
casey, line 94:
```suggestion
                  "Opening an issue on https://github.com/soenkehahn/cradle/issues ",
```

src/error.rs
```diff
@@ -70,24 +82,35 @@ impl Display for Error {
...
+                  "including the following information:",
```
casey, line 95:
```suggestion
                  "with the following information:",
```

casey (commented):
Lots of minor comments, but looks good overall.

memory-test/src/bin/run_test.rs
```diff
@@ -0,0 +1,56 @@
+use anyhow::Result;
+use cradle::prelude::*;
+use std::process::{Command, Stdio};
+
+fn from_mb(mega_bytes: usize) -> usize {
```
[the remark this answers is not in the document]
soenkehahn: No, good catch!

memory-test/src/bin/run_test.rs
```diff
@@ -0,0 +1,56 @@
...
+    Ok(bytes)
+}
+
+#[rustversion::attr(since(1.48), allow(clippy::manual_strip))]
+fn strip_prefix<'a>(string: &'a str, prefix: &'a str) -> &'a str {
+    if string.starts_with(prefix) {
+        &string[prefix.len()..]
+    } else {
+        panic!("{} doesn't start with {}", string, prefix);
+    }
```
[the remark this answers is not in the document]
soenkehahn: I think you mean `strip_prefix`. That is not stable on rust `1.41`. :|

src/error.rs
soenkehahn: I went with this: 1449f40

src/error.rs
soenkehahn: Done.

src/collected_output.rs
soenkehahn: Maybe, yes. That shouldn't be done in this PR though.

memory-test/src/bin/produce_bytes.rs
```diff
@@ -0,0 +1,21 @@
...
+    let mut stdout = stdout();
+    while bytes > 0 {
+        if bytes >= buffer.len() {
+            stdout.write_all(buffer)?;
+            bytes -= buffer.len();
+        } else {
+            stdout.write_all(&[b'x'])?;
+            bytes -= 1;
+        }
+    }
```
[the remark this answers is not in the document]
soenkehahn: Ooh, very good.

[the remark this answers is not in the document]
soenkehahn: Ok, it wasn't just as simple but this works: 13dd24e

.github/workflows/ci.yaml
```diff
@@ -73,6 +73,7 @@ jobs:
       - name: Install just
```
casey, line 73: If you want to get a specific version of just, you can do `curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to DEST --tag VERSION`

memory-test/src/bin/run_test.rs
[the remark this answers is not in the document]
casey: Ahh, gotcha, I didn't realize that.

.github/workflows/ci.yaml
soenkehahn: Does that work on windows?

casey: Uhhhhhhh, I don't actually know 😅

soenkehahn: Yeah, no, it doesn't. :|

soenkehahn: I've opened this issue to explain what the actual problems are: https://github.com/casey/just/issues/943. I.e. I'm not at all interested in getting a specific `just` version, the newest would be just fine. I just want a way that works on github CI on all platforms.

PR #144, opened 2021-08-06
Add Debug instances for every type consistently

soenkehahn:


casey:
We should also deny the lint for this too:

```rust
#![deny(missing_debug_implementations)] 
```

Other than that, looks good.

PR #148, opened 2021-08-10
Don't read stderr into memory when not captured

### LINKED ISSUE #26: Don't collect `stdout` and `stderr` into memory when not needed

soenkehahn:
Similar to #143. Fixes #26.


memory-tests/src/bin/produce_bytes.rs
```diff
@@ -0,0 +1,22 @@
...
+    let buffer = &[b'x'; 1024];
+    let mut stream: Box<dyn Write> = match stream_type.as_str() {
+        "stdout" => Box::new(io::stdout()),
+        "stderr" => Box::new(io::stderr()),
+        _ => panic!("unknown stream type: {}", stream_type),
+    };
+    while bytes > 0 {
+        let chunk_size = bytes.min(1024);
+        stream.write_all(&buffer[..chunk_size])?;
+        bytes -= chunk_size;
```
casey, line 18: Just came up with this marginally more elegant formulation:

```suggestion
        let chunk = &buffer[..bytes];
        stream.write_all(chunk)?;
        bytes -= chunk.len();
```

src/collected_output.rs
```diff
@@ -54,15 +54,21 @@ impl Waiter {
         });
         let mut context_clone = context.clone();
         let capture_stderr = config.capture_stderr;
-        let stderr_join_handle = thread::spawn(move || -> io::Result<Vec<u8>> {
-            let mut collected_stderr = Vec::new();
+        let stderr_join_handle = thread::spawn(move || -> io::Result<Option<Vec<u8>>> {
```
casey, line 57: I wonder if this could be DRYed up. The code here seems nontrivial, so merging the stdout and stderr capture would be nice.

casey (commented):
Looks good to me, minor comment about deduplicating the code, but if you can't see a good way to do it I don't think it should hold this PR up.

memory-tests/src/bin/produce_bytes.rs
soenkehahn: This doesn't work. `buffer[..bytes]` panics when `bytes` is bigger than `buffer.len()`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=69638e3fc596ebf7cc24ec7e4450a121

src/collected_output.rs
soenkehahn: What do you think of 4015e1c?

memory-tests/src/bin/produce_bytes.rs
casey: Oh damn, I wonder why I thought that 😅 Maybe from python.

src/collected_output.rs
casey: Nice, looks good.

PR #149, opened 2021-08-10
Install just through github action

soenkehahn:


casey:
Sweet, looks good to me! Awesome that it even works on windows.

PR #150, opened 2021-08-10
Add hint to FileNotFoundWhenExecuting in case of whitespace in executable names

soenkehahn:


casey:
This is a great hint. I think the square brackets are a little confusing though. I think the ideal message would be:

```
Did you mean to run `does`, with `not` and `exist` as arguments?
```

(I also like backticks for setting off literal text, but TIOLI.)

soenkehahn:
How would this look like with only one argument? `Did you mean to run 'foo', with 'bar' as argument?`?
And how about three arguments? `Did you mean to run 'foo', with 'bar', 'baz' and 'huhu' as arguments?`?
I felt this was a bit overly complicated, so I went with square brackets. Since they're used in lots of languages for list syntax. But maybe sticking with English is better?

Also, I'm using single quotes everywhere else in the library, so I'm sticking with that for this PR.

casey:
> How would this look like with only one argument? `Did you mean to run 'foo', with 'bar' as argument?`?
> And how about three arguments? `Did you mean to run 'foo', with 'bar', 'baz' and 'huhu' as arguments?`?

Yeah, something like that. I have some [weird](https://github.com/casey/just/blob/master/src/list.rs) [helpers](https://github.com/casey/just/blob/master/src/list.rs) that do this for `just`.

> I felt this was a bit overly complicated, so I went with square brackets. Since they're used in lots of languages for list syntax. But maybe sticking with English is better?

I definitely agree it's complicated. `[` and `]` are used for list syntax, but this isn't a context that I'm really used to seeing them. An alternative that might be more readable is something like this: `Did you mean to run 'does', with the arguments ['not', 'exist']?` I see the word "arguments" before the list, so I'm sort of primed to recognize it as a list of arguments.

> Also, I'm using single quotes everywhere else in the library, so I'm sticking with that for this PR.

Fair enough!

casey:
Nice, looks good.

PR #154, opened 2021-08-12
Move everything out of the root module into submodules

soenkehahn:


casey:
Looks good to me! One thing PRs like this make me want is integration tests, which, since they're subject to the same visibility rules as users, would break if public visibility of symbols changed.

soenkehahn:
We do have integration tests: https://github.com/soenkehahn/cradle/blob/a68d64e792fb8b67dd7a4ea083d5c6abfc13f563/tests/integration.rs. They don't systematically test all things from `input` or `output` though. And they don't check exports from the root module of the crate.

Thanks for the review!

PR #156, opened 2021-08-13
[CI] Disable cache

soenkehahn:
It seems like caching creates some problems on CI, so I think we should disable it, at least for now. `cradle` doesn't have that many dependencies anyways.


soenkehahn:
Nevermind, something else seems to be going on on mac (on CI).

soenkehahn:
Closing in favor of #158.

PR #158, opened 2021-08-13
Fix CI builds for new versions of dependencies from crates.io

soenkehahn:
There's a weird problem caused be the newly released `bitflags-1.3.1` on CI on older rustcs. This causes `cargo build --bin test_executables_helper --features test_executables` to fail. Unfortunately we only execute that command from within the test-suite. And if it fails, the test-suite hangs. :( This PR makes the test-suite crash when this happens.

Edit: point to bitflags instead of libc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ version = "1.0.0"
 optional = true
 
 [target.'cfg(unix)'.dependencies.nix]
-version = "=0.21.0"
+version = "0.21.0"
 optional = true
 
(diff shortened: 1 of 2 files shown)


soenkehahn:
Here's an issue about the root cause: https://github.com/nix-rust/nix/issues/1491.

This PR now also pins `bitflags` to version `1.2.1` to work around this issue. Since `bitflags` is only a dev-dependency, I think that should be fine.

PR #160, opened 2021-08-13
Don't use cmd! in error messages

soenkehahn:


soenkehahn:
Closing in favor of #161.

PR #161, opened 2021-08-13
Rename main macros and Input methods

### LINKED ISSUE #119: Consider alternative names to `cmd!`, `cmd_unit!`, and `cmd_result!`
This is totally subjective, so feel free to close this if you don't agree, but I kind of think the existing macro names aren't the best.

I think `cmd_unit` is actually extremely common, just judging from our own scripts, and it would be nice if it had a nicer, shorter name.

I like the names `run!(…)`, `output!(…)`, and `result!(…)`. `run` is the default for when you don't want anything, and `output` and `result` tell you what they're going to return.

Also remember to rename any instances of `cmd!` and friends in error messages, which should also probably not use the macro `!`, since these error messages can now be elicited by methods.

soenkehahn:
Fixes #119.

This PR renames the main macros and the `Input::run*` methods. It also makes some other changes that seemed necessary given these renamings. To make it easier to review the PR I took care to have the single commits make sense:
- All commits where the message starts with 'Rename ...' are simple renamings, and contain no other changes.
- The other commits contain more interesting changes.

So maybe it makes sense to review the PR commit by commit. Or maybe not. Not sure how well github handles comments on commits...

Notes:

Commit https://github.com/soenkehahn/cradle/pull/161/commits/2609b633a080c543b178bfba8e7a623cf9468cf2 changes the order of the documentation. I thought, since `run!` (former `cmd_unit!`) is now shorter it would be more natural to document it first. That also allows us to first introduce input polymorphism and then -- when talking about `run_output!` (former `cmd!`) -- introduce return-type polymorphism. Which makes for a slightly nicer difficulty curve.

Commit https://github.com/soenkehahn/cradle/pull/161/commits/2fab2ee75a358e3bac21e4d2e9622050a451d43c renames the `RunResult` type to `ChildOutput`. And consequently it renames the containing module from `run_result` to `child_output`. This is because it was a bit confusing that we had 1. a macro called `run_result`, 2. an `Input::run_result` method with the same behavior and then 3. a `run_result` module that is pretty unrelated to the other items with the same name.

### DIFF
--- a/examples/readme.rs
+++ b/examples/readme.rs
@@ -1,8 +1,9 @@
 use cradle::prelude::*;
 
 fn main() {
-    let StdoutTrimmed(git_version) = run_output!(%"git --version");
-    eprintln!("git version: {}", git_version);
+    // output git version
+    run!(%"git --version");
+    // output configured git user
     let (StdoutTrimmed(git_user), Status(status)) = run_output!(%"git config --get user.name");
     if status.success() {
         eprintln!("git user: {}", git_user);
--- a/examples/readme.rs
+++ b/examples/readme.rs
@@ -1,9 +1,9 @@
 use cradle::prelude::*;
 
 fn main() {
-    let StdoutTrimmed(git_version) = cmd!(%"git --version");
+    let StdoutTrimmed(git_version) = run_output!(%"git --version");
     eprintln!("git version: {}", git_version);
-    let (StdoutTrimmed(git_user), Status(status)) = cmd!(%"git config --get user.name");
+    let (StdoutTrimmed(git_user), Status(status)) = run_output!(%"git config --get user.name");
     if status.success() {
         eprintln!("git user: {}", git_user);
     } else {
--- a/src/error.rs
+++ b/src/error.rs
@@ -55,7 +55,7 @@ impl Error {
 pub fn panic_on_error<T>(result: Result<T, Error>) -> T {
     match result {
         Ok(t) => t,
-        Err(error) => panic!("cmd!: {}", error),
+        Err(error) => panic!("run_output!: {}", error),
     }
 }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,4 @@
-//! The [`Error`] type used in the return type of [`cmd_result!`].
+//! The [`Error`] type used in the return type of [`run_result!`].
 
 use crate::config::Config;
 use std::{ffi::OsString, fmt::Display, io, process::ExitStatus, string::FromUtf8Error, sync::Arc};
@@ -173,7 +173,7 @@ mod tests {
 
     #[test]
     fn invalid_utf8_to_stdout_has_source() {
-        let result: Result<StdoutUntrimmed, crate::Error> = cmd_result!(
+        let result: Result<StdoutUntrimmed, crate::Error> = run_result!(
             executable_path("cradle_test_helper").to_str().unwrap(),
             "invalid utf-8 stdout"
         );
@@ -182,7 +182,7 @@ mod tests {
 
     #[test]
     fn invalid_utf8_to_stderr_has_source() {
-        let result: Result<Stderr, crate::Error> = cmd_result!(
+        let result: Result<Stderr, crate::Error> = run_result!(
             executable_path("cradle_test_helper").to_str().unwrap(),
             "invalid utf-8 stderr"
         );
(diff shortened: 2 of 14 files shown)


src/error.rs
```diff
@@ -55,7 +55,7 @@ impl Error {
 pub fn panic_on_error<T>(result: Result<T, Error>) -> T {
     match result {
         Ok(t) => t,
-        Err(error) => panic!("cmd!: {}", error),
+        Err(error) => panic!("cradle error: {}", error),
```
soenkehahn, line 58: Instead of referencing a specific macro, I changed this to something generic.

examples/readme.rs
```diff
@@ -1,9 +1,10 @@
 use cradle::prelude::*;
 
 fn main() {
-    let StdoutTrimmed(git_version) = cmd!(%"git --version");
-    eprintln!("git version: {}", git_version);
-    let (StdoutTrimmed(git_user), Status(status)) = cmd!(%"git config --get user.name");
+    // output git version
```
casey, line 4: Unrelated to this PR: Why is this code in the readme and in an example? Could it just live in the readme?

src/input.rs
```diff
@@ -1,5 +1,4 @@
-//! The [`Input`] trait that defines all possible inputs to [`cmd!`],
-//! [`cmd_unit!`] and [`cmd_result!`].
+//! The [`Input`] trait that defines all possible inputs to `cradle`.
```
casey, line 1:
```suggestion
//! The [`Input`] trait that defines all possible inputs to a command.
```

TÌOLI

src/input.rs
```diff
@@ -350,13 +349,13 @@ tuple_impl!(0, A, 1, B, 2, C, 3, D, 4, E,);
 tuple_impl!(0, A, 1, B, 2, C, 3, D, 4, E, 5, F,);
 tuple_impl!(0, A, 1, B, 2, C, 3, D, 4, E, 5, F, 6, G,);
 
-/// All elements of the given [`Vec`] are used as arguments to [`cmd!`].
+/// All elements of the given [`Vec`] are used as arguments to the child process.
```
casey, line 352:
```suggestion
/// All elements of the given [`Vec`] are used as arguments to the command.
```

TIÖLI

tests/integration.rs
```diff
@@ -125,12 +125,12 @@ fn user_supplied_errors_succeeding() {
 
     #[derive(Debug)]
     enum Error {
-        CmdError(cradle::Error),
+        CradleError(cradle::Error),
```
casey, line 138: Maybe just `Cradle`, since it's already an `Error` enum?

casey (commented):
Nice, this looks great. A couple minor comments, but all TIOLI. 

Splitting it up made it super easy to review, so thanks for that. There are prev/next buttons on each commit page, so you can just start at the beginning and click next until you get to the last commit, which is nice.

examples/readme.rs
soenkehahn: It's in an example rust file so that it can be type-checked and run on CI. And then it's rendered into the readme through this recipe: https://github.com/soenkehahn/cradle/blob/639f3ede9b8ab1e72a2e05d208835c49111429e9/justfile#L27..L28. So that I don't have to manually copy that. Also, there's a CI check that makes sure that the `readme.rs` and the `README.md` are in sync: https://github.com/soenkehahn/cradle/blob/639f3ede9b8ab1e72a2e05d208835c49111429e9/justfile#L30..L32. (And then I have a local git pre-commit hook that calls the first recipe.)

This is basically trying to replicate the nice features of doctests for the readme. I wish there was a better tool for this, e.g. `cargo` could support that in some way. But I haven't found one.

src/input.rs
soenkehahn: I think 'command' is a bit weird for referring to child processes. Outside of Rust's standard library, I never think of a child process as a 'command'. And just using the word 'command' in lower-case, in natural English doesn't necessarily make me think of `std::process::Command`. (And -- strictly speaking -- `std::process::Command` is not a child process, but just the configuration for how to spawn one.)

But yeah, using `cradle` here is also not so nice. What do you think about saying '... inputs to a child process'?

casey: I think that sounds good.

tests/integration.rs
soenkehahn: Yes, done.

examples/readme.rs
casey: One thing that might work is including the readme as a doc comment somewhere with `#[doc = include_str!("../README.md")]`. The rust codeblocks would then be compiled as doc tests.

PR #164, opened 2021-08-15
Move Sized constraint from Input methods to trait

soenkehahn:
This technically adds a `Sized` constraint to `Input::configure` that didn't exist before. All of our inputs implement `Sized` though.

Update: Even though `configure` didn't explicitly have a `Sized` constraint, you still couldn't implement it without, since it takes `self`.


casey:
LGTM

PR #166, opened 2021-08-17
[Refactoring] Implement the main macros in terms of Input::methods on tuples

soenkehahn:
This PR implements the macros `run!`, `run_output!` and `run_result!` in terms of the methods on `Input` and in turn implements the methods on `Input` in terms of the functionality from `child_output`. And moves some code around to hopefully make more sense. This also means that some internal functions don't have to be public anymore, because they won't show up in code generated by the macros. Therefore, this is technically a breaking change. All the items that were changed to non-public now were `#[doc(hidden)]` before.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -268,7 +268,7 @@ impl Input for String {
 /// ```
 ///
 /// [`split_whitespace`]: str::split_whitespace
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct Split<T: AsRef<str>>(pub T);
 
 impl<T: AsRef<str>> Input for crate::input::Split<T> {
--- a/src/input.rs
+++ b/src/input.rs
@@ -109,6 +109,7 @@ pub trait Input: Sized {
     ///
     /// ("touch", "foo").run();
     /// ```
+    #[track_caller]
     fn run(self) {
         let () = self.run_output();
     }
@@ -122,6 +123,7 @@ pub trait Input: Sized {
     /// let StdoutTrimmed(output) = ("echo", "foo").run_output();
     /// assert_eq!(output, "foo");
     /// ```
+    #[track_caller]
     fn run_output<O>(self) -> O
     where
         O: Output,
--- a/src/input.rs
+++ b/src/input.rs
@@ -109,7 +109,7 @@ pub trait Input: Sized {
     ///
     /// ("touch", "foo").run();
     /// ```
-    #[track_caller]
+    #[rustversion::attr(since(1.46), track_caller)]
     fn run(self) {
         let () = self.run_output();
     }
@@ -123,7 +123,7 @@ pub trait Input: Sized {
     /// let StdoutTrimmed(output) = ("echo", "foo").run_output();
     /// assert_eq!(output, "foo");
     /// ```
-    #[track_caller]
+    #[rustversion::attr(since(1.46), track_caller)]
     fn run_output<O>(self) -> O
     where
         O: Output,
--- a/src/input.rs
+++ b/src/input.rs
@@ -151,19 +151,32 @@ pub trait Input: Sized {
         O: Output,
     {
         let context = Context::production();
-        run_result_with_context(self, context)
+        run_result_with_context(context, self)
     }
 }
 
-pub(crate) fn run_result_with_context<I, O, Stdout, Stderr>(
-    input: I,
+#[cfg(test)]
+pub(crate) fn run_result_with_context_unit<Stdout, Stderr, I>(
     context: Context<Stdout, Stderr>,
-) -> Result<O, Error>
+    input: I,
+) -> Result<(), Error>
 where
-    I: Input,
-    O: Output,
     Stdout: Write + Clone + Send + 'static,
     Stderr: Write + Clone + Send + 'static,
+    I: Input,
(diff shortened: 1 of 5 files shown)


src/input.rs
```diff
@@ -102,8 +109,9 @@ pub trait Input: Sized {
     ///
     /// ("touch", "foo").run();
     /// ```
+    #[track_caller]
```
soenkehahn, line 112: It's probably a bit unconventional to have `#[track_caller]` in public functions of a library. But since `Input::run` (and `Input::run_output`) are designed to panic, I think it's probably best to have the panic source location match the **call-site** of these methods.

src/input.rs
```diff
@@ -141,10 +150,39 @@ pub trait Input: Sized {
     where
         O: Output,
     {
-        crate::run_result!(self)
+        let context = Context::production();
+        run_result_with_context(context, self)
     }
 }
 
+pub(crate) fn run_result_with_context<Stdout, Stderr, I, O>(
```
soenkehahn, line 158: This (and `run_result_with_context_unit`) seem like they should be methods on `Input`. But then they inherit the `pub` visibility from that trait. So I made them `pub(crate)` standalone functions.

src/input.rs
```diff
@@ -246,7 +284,7 @@ impl Input for String {
 /// ```
 ///
 /// [`split_whitespace`]: str::split_whitespace
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
soenkehahn, line 287: This is needed for the `tuple_up!` tests.

src/lib.rs
```diff
@@ -311,14 +313,6 @@ mod tests {
         test_executable("test_executables_helper")
     }
 
-    macro_rules! run_result_with_context_unit {
```
soenkehahn, line 314: This is now a function in the `input` module.

src/lib.rs
```diff
@@ -608,9 +602,10 @@ mod tests {
         #[test]
         fn vector_of_non_strings() {
             let context = Context::test();
-            let config: Vec<LogCommand> = vec![LogCommand];
+            let log_commands: Vec<LogCommand> = vec![LogCommand];
             let StdoutTrimmed(stdout) =
-                crate::run_result_with_context!(context.clone(), config, %"echo foo").unwrap();
+                run_result_with_context(context.clone(), (log_commands, Split("echo foo")))
+                    .unwrap();
```
soenkehahn, line 608: A lot of these tests changed from using the macro interface to using a tuple based interface through `run_result_with_context` and `run_result_with_context_unit`.

src/lib.rs
```diff
@@ -1448,7 +1439,7 @@ mod tests {
             let unused_key = find_unused_environment_variable();
             env::set_var(&unused_key, "foo");
             let StdoutTrimmed(output) =
-                run_output!(test_helper(), "echo", &unused_key, Env(unused_key, "bar"));
+                run_output!(test_helper(), "echo", &unused_key, Env(&unused_key, "bar"));
```
soenkehahn, line 1442: This is now needed because `run_output!` constructs a tuple of all input values.

src/macros.rs
```diff
@@ -56,36 +55,76 @@ macro_rules! run_output {
...
     ($($args:tt)*) => {{
-        let context = $crate::context::Context::production();
-        $crate::run_result_with_context!(context, $($args)*)
+        $crate::input::Input::run_result($crate::tuple_up!($($args)*))
     }}
 }
 
 #[doc(hidden)]
 #[macro_export]
-macro_rules! run_result_with_context {
```
soenkehahn, line 66: This is now a non-public function in module `input`.

src/macros.rs
```diff
@@ -56,36 +55,76 @@ macro_rules! run_output {
...
-        $crate::child_output::ChildOutput::run_child_process_output($context, config)
-    }}
-}
-
-#[doc(hidden)]
-#[macro_export]
-macro_rules! configure {
-    (config: $config:ident, args: % $head:expr $(,)?) => {
-        $crate::input::Input::configure($crate::input::Split($head), &mut $config);
+macro_rules! tuple_up {
```
soenkehahn, line 64: The diff for `tuple_up!` looks pretty bad. Probably easier to look at this: https://github.com/soenkehahn/cradle/blob/macros-in-terms-of-tuples/src/macros.rs#L62..L130

src/lib.rs
```diff
@@ -750,17 +747,16 @@ mod tests {
...
-            run_result_with_context_unit!(context.clone(), %"echo foo").unwrap();
+            run_result_with_context_unit(context.clone(), Split("echo foo")).unwrap();
             assert_eq!(context.stdout(), "foo\n");
         }
 
         #[test]
         fn relays_stdout_for_non_zero_exit_codes() {
             let context = Context::test();
-            let _: Result<(), Error> = crate::run_result_with_context!(
+            let _ = run_result_with_context_unit(
```
casey, line 757:
```suggestion
            un_result_with_context_unit(
```

casey:
Sweet, looks good to me. I tried improving the tuple_up macro by adding a base case: 

```rust
macro_rules! tuple_up {
    () => {()};
    …
}
```

But that didn't work.

PR #174, opened 2021-08-31
Remove comment about api instability

soenkehahn:
I do expect normal API churn from now on, no drastic changes. So I think it'd be ok to remove this warning. Especially since the versions still start with `0.0.`.

One thing that may happen soon is to allow users to write their own `Input` and `Output` impls. But that should only change parts of the API that are currently `#[doc(hidden)]`.


casey:
LGTM. Cradle is now 100% stable enterprise-ready software that you can feel confident shipping to prod.

PR #175, opened 2021-08-31
Allow to implement custom impls for `Input`

soenkehahn:
One open question: Currently the `configure` methods in the `Input` impls are all `#[doc(hidden)]`. Which makes some sense, since users are usually not going to call those. But since the method is unhidden in the trait now, maybe we should also unhide the impl methods?

Update: I just tried to unhide `configure` in the impls and it causes the doc comments for the impls to be collapsed. So you have to click a little `+` to see them (and the configure method). So I think that would be worse.

### DIFF
--- a/justfile
+++ b/justfile
@@ -33,6 +33,6 @@ render-readme-check:
 
 forbidden-words:
   ! grep -rni \
-    'dbg!\|fixme\|todo\|ignore' \
+    'dbg!\|fixme\|todo\|#\[ignore\]' \
     src tests examples
   @echo No forbidden words found
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,7 +2,7 @@
 
 use std::{ffi::OsString, path::PathBuf, sync::Arc};
 
-/// Internal type that configures how to run a child process.
+/// Used by `Input` implementations to configure how child processes are run.
 /// Usually you don't have to use this type directly.
 ///
 /// See also [Custom `Input` impls](crate::Input#custom-input-impls).
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,7 +2,10 @@
 
 use std::{ffi::OsString, path::PathBuf, sync::Arc};
 
-#[doc(hidden)]
+/// Internal type that configures how to run a child process.
+/// Usually you don't have to use this type directly.
+///
+/// See also [Custom `Input` impls](crate::Input#custom-input-impls).
 #[rustversion::attr(since(1.48), allow(clippy::rc_buffer))]
 #[derive(Debug, Clone)]
 pub struct Config {
--- a/src/input.rs
+++ b/src/input.rs
@@ -101,31 +101,39 @@ use std::{
 /// The provided `Input` implementations should be sufficient for most use cases,
 /// but custom `Input` implementations can be written to extend `cradle`.
 ///
-/// Here's an example of an `InContainer` struct, that makes it easy to
-/// run commands in a [`podman`](https://podman.io/) container:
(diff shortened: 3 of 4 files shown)


justfile
```diff
@@ -33,6 +33,6 @@ render-readme-check:
 
 forbidden-words:
   ! grep -rni \
-    'dbg!\|fixme\|todo\|ignore' \
+    'dbg!\|fixme\|todo\|#\[ignore\]' \
```
soenkehahn, line 36: Needed because the word 'ignore' appears in a comment now.

src/config.rs
```diff
@@ -2,7 +2,10 @@
 
 use std::{ffi::OsString, path::PathBuf, sync::Arc};
 
-#[doc(hidden)]
+/// Internal type that configures how to run a child process.
```
casey, line 5:
```suggestion
/// Used by `Input` implementations to configure how child processes are run.
```

src/input.rs
```diff
@@ -95,8 +95,60 @@ use std::{
 /// let hex = to_hex((Stdin(&[14, 15, 16]), Stdin(&[17, 18, 19])));
 /// assert_eq!(hex, "0E0F10111213");
 /// ```
+///
+/// ## Custom [`Input`] impls
+///
+/// For most use-cases, it's not needed to write your own impls
+/// for the [`Input`] trait.
+/// But sometimes it can be helpful, e.g. when you're writing helper
+/// functions around `cradle`.
```
casey, line 104:
```suggestion
/// The provided `Input` implementations should be sufficient for most use cases,
/// but custom `Input` implementations can be written to extend `cradle`.
```

src/input.rs
```diff
@@ -95,8 +95,60 @@ use std::{
...
+/// Here's an example of an `InContainer` struct, that makes it easy to
```
casey, line 105:
```suggestion
///
/// Here's an example of an `InContainer` struct, that makes it easy to
```

casey (commented):
Some random wording comments, but otherwise looks good.

src/input.rs
```diff
@@ -95,8 +95,60 @@ use std::{
...
+/// # #[cfg(target_os = "linux")]
+/// # {
+/// use cradle::prelude::*;
+/// use cradle::config::Config;
+///
+/// struct InContainer;
+///
+/// impl Input for InContainer {
+///     fn configure(self, config: &mut Config) {
+///         (LogCommand, Split("podman run alpine:3.14")).configure(config);
```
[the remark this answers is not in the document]
soenkehahn: Yeah... I think if I wanted to implement this for real, I would rather do something like:
```rust
fn in_container<I: Input, O: Output>(input: I) -> O {
    run_output!(LogCommand, %"podman run alpine:3.14", input)
}
```
And that would avoid the issue that you have to make sure to use `InContainer` as the first argument.
And -- given that I wouldn't even recommend doing this -- maybe we should have a better example? I'll try to come up with something.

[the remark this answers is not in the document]
soenkehahn: Just pushed a new example: https://github.com/soenkehahn/cradle/pull/175/commits/5015c9d49ad6bd7cb039dbd6f411309c2f4f61dc. Could you take another look, please?

casey:
The new example is simple and concrete, but the motivation isn't clear to me. The podman example had a gotcha, but it was clearer why it would be useful. But maybe the new example is fine, since it does demonstrate how to write custom `Input` implementations.

src/input.rs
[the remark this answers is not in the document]
soenkehahn: Pushed yet another commit to change the example.

casey:
Nice, LGTM

PR #177, opened 2021-09-01
Rename from_run_result to from_child_output

soenkehahn:
`ChildOutput` was once called `RunResult`, but this trait method wasn't renamed when the type was renamed.


casey:
LGTM

PR #178, opened 2021-09-01
Simplify Error type

soenkehahn:
`Error` needed to implement `Clone` at some point, but it doesn't anymore. So it can be simplified a bit.


casey:
LGTM

PR #183, opened 2021-09-07
Improve documentation for `Output`

soenkehahn:
I tried to come up with a good example for a custom impl for the `Output` trait, but that turned out to be a bit more involved. I also realized that when using `cradle` I never really found myself in a position where I wanted to be able to write a custom `Output` impl. So the examples that I came up with were rather made-up. (I often wanted a new `Output` impl within `cradle`, but that doesn't require a nice api to write `Output` impls outside of it.) So I put this PR together, that at least documents that this is currently not really recommended. The documentation also links to #184, which invites people to speak up when they find use-cases for this.

### DIFF
--- a/src/child_output.rs
+++ b/src/child_output.rs
@@ -11,6 +11,9 @@ use std::{
 
 /// Internal type to capture all the outputs of a child process.
 /// Usually you don't have to use this type directly.
+///
+/// See also the documentation for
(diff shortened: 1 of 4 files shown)


casey:
Reworded a comment a bit (I like avoiding lists when possible, and just using prose) but otherwise looks good.

PR #188, opened 2021-09-10
Add design philosophy section to readme

soenkehahn:
I was wondering whether to add this to the readme or to the main documentation (in `lib.rs`). Since the readme will be rendered into the crates' pages on crates.io, I think adding this to the readme is best.


README.md
```diff
@@ -24,5 +26,17 @@ fn main() {
...
+
+`cradle` is meant to make it as easy as possible to run child processes,
+while making it very hard to use incorrectly.
+As such it provides an interface that is very concise, yet flexible,
+but tries to avoid any behavior that would be unexpected or surprising.
+
+`cradle` decidedly does _not_ try to emulate any syntax of `bash` or other shells,
+like piping (`|`) or shell expansion (e.g. globs, like `*`).
+Instead, it is aiming to be a convenience wrapper around your
+operating system's interface for running child processes.
```
casey, line 39:
```suggestion
`cradle` does not try to emulate the syntax or functionality of `bash` or other shells,
such as pipes (`|`), globs (`*`), or other string expansion.
Instead, it aims to be a convenient wrapper around the
operating system's interface for running child processes.
```

casey (commented):
I see you snuck in a `.php` file!

Is that to make sure that the example in the readme compiles? If it isn't necessary to run it as a standalone example, then there might be other ways to accomplish this, for example by including the readme in a doc-comment, a.la. `#[doc = include_str!("../README.md")]`

although on second thought, that might not be supported on the MSRV.

In any case, minor wording changes, but otherwise looks good.

soenkehahn:
> I see you snuck in a `.php` file!
> 
> Is that to make sure that the example in the readme compiles?

Yes.

> If it isn't necessary to run it as a standalone example, then there might be other ways to accomplish this, for example by including the readme in a doc-comment, a.la. `#[doc = include_str!("../README.md")]`

If this were supported from the MSRV upwards, that sounds like it could be nice. Although the current solution works pretty well, and I don't see a super compelling reason to change it. :shrug:

PR #191, opened 2021-09-20
Allow only `&'static str` in `Split`

soenkehahn:
This narrows down the types allowed as arguments to `Split` (and therefore the types of values after `%` in macros). The reasoning is that
- this has hardly any downsides, since `Split` is meant as a convenience for working with string literals,
- it prevents surprising behavior in some corner-cases, for an example see https://github.com/matklad/cradle-of-tar.

For more discussion, see https://www.reddit.com/r/rust/comments/poqlo9/cradle_run_child_processes_with_ease/hd00yla/?utm_source=reddit&utm_medium=web2x&context=3.

Do we have any reasonable usecases where arguments to `Split` are not `&'static str`?

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -329,9 +329,6 @@ impl Input for String {
 ///
 /// let StdoutTrimmed(output) = run_output!(Split("echo foo"));
 /// assert_eq!(output, "foo");
-///
-/// let StdoutTrimmed(output) = run_output!(Split(format!("echo {}", 100)));
-/// assert_eq!(output, "100");
 /// ```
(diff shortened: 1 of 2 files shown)


casey:
I think this is reasonable. I can't think of any really important use cases for splitting non-static strings, and if someone has a use case, we can easily re-enable it.

casey:
We should probably also only provide `Input` implementations for `SplitWhitespace<'static>`, `Split<'static>`, and `SplitAsciiWhitespaceCopy<'static>`.

PR #192, opened 2021-09-22
Exclude .github directory from crate

olivierlemasle:


soenkehahn:
Thanks for opening the PR. IIUC this will exclude the `.github` directory from the release tarballs. Or does this have any other effect? Why do you want the directory to be excluded?

olivierlemasle:
Hi @soenkehahn, yes, this simply excludes the  `.github` directory from the crate tarball. This directory is unnecessary to the crate's consumers.

Actually, I opened this PR because I'm packaging `cradle` for Fedora (as par of `just` packaging), and Fedora recommends excluding the CI scripts: cf https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_excluding_unnecessary_files

soenkehahn:
It's great that you're packaging `cradle` for Fedora, thanks a lot for that. I'll merge the PR and the change will take effect in the next release.

Out of curiosity: Is the source tarball going to be included in a Fedora package? Or is it just going to be referenced by url somehow?

PR #200, opened 2021-12-25
Derive Copy for input::LogCommand

soenkehahn:
I ran into an ownership error message with `LogCommand` in another project. It can be easily worked around by using `&LogCommand`, but I thought deriving `Copy` also is a good idea.

@casey: Do you know whether this would be a minor or a major version bump? (I wouldn't necessarily cut a release for this though.)


casey:
I think this is just a minor version bump, since it's like adding a function.

LGTM!

PR #201, opened 2021-12-30
Provide `Input` and `Output` impls for tuples up to size 16

soenkehahn:
I started using `Split` (or `%`) less and started using tuples more and ran into this limit.


casey:
LGTM!

https://twitter.com/rodarmor/status/1465212523196530690/photo/1

PR #204, opened 2022-05-02
Add new rustc versions to ci

soenkehahn:


src/input.rs
```diff
@@ -495,7 +495,16 @@ where
 {
     #[doc(hidden)]
     fn configure(self, config: &mut Config) {
-        for t in std::array::IntoIter::new(self) {
+        #[rustversion::all(since(1.51), before(1.59))]
+        fn array_to_iter<T, const N: usize>(array: [T; N]) -> impl Iterator<Item = T> {
+            std::array::IntoIter::new(array)
```
soenkehahn, line 500: This function is deprecated since `1.59`. But `IntoIterator::into_iter` is not available in e.g. `1.51` yet.

casey (approved):

PR #206, opened 2022-05-02
Add just recipe to run tests for all supported rustc versions locally

soenkehahn:


justfile
```diff
@@ -36,3 +36,20 @@ forbidden-words:
...
+
+  export RUSTFLAGS="--deny warnings"
+
+  # install yq with: pip3 install yq
+  versions=$(cat .github/workflows/ci.yaml \
+    | yq -r '.jobs.all.strategy.matrix.rust | sort | join(" ")' \
+    ;)
+  for version in $versions
+  do
+    rustup override set $version
```
casey, line 52: If the recipe fails, will the override still be set?

soenkehahn: Yes. I thought there was no easy way to avoid that, but then I found [`RUSTUP_TOOLCHAIN`](https://rust-lang.github.io/rustup/environment-variables.html). Good catch! Fixed in 3bfc22ad.

casey: Nice, LGTM

PR #207, opened 2022-05-02
Don't fork a thread for stdin if not necessary

soenkehahn:


src/collected_output.rs
```diff
@@ -52,11 +52,13 @@ impl Waiter {
         Stdout: Write + Send + Clone + 'static,
         Stderr: Write + Send + Clone + 'static,
     {
-        let config_stdin = config.stdin.clone();
-        let stdin_join_handle = thread::spawn(move || -> io::Result<()> {
-            child_stdin.write_all(&config_stdin)?;
-            Ok(())
-        });
+        let stdin_join_handle = match config.stdin.clone() {
```
casey, line 55: It looks like this can be done with `.map`. TIOLI.

casey:
LGTM!

src/collected_output.rs
soenkehahn: Yeah, I actually tried that out but thought that it obscured a bit what was going on.

soenkehahn: Some older versions of `clippy` (e.g. in rustc `1.53`) also complain about this. So I'm switching to `.map`. If you *and* `clippy` say so... :shrug:

PR #209, opened 2022-05-04
Refactor panic source location tests

soenkehahn:
This PR gets rid of the `test_executables_panic` executable, while re-implementing the same tests in a different way. I think this is better since the executable is not meant for users of the library and as such ideally shouldn't show up in the `Cargo.toml` file.

(I also would like to get rid of `test_executables_helper` at some point.)

### DIFF
--- a/tests/panic_source_locations.rs
+++ b/tests/panic_source_locations.rs
@@ -1,7 +1,6 @@
-#[allow(unused_imports)]
+#[rustversion::since(1.46)]
 use cradle::prelude::*;
-#[allow(unused_imports)]
-use pretty_assertions::assert_eq;
+#[rustversion::since(1.46)]
 use std::{
     panic::{set_hook, take_hook},
     sync::{Arc, Mutex},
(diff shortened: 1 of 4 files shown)


tests/panic_source_locations.rs
```diff
@@ -0,0 +1,39 @@
+#[allow(unused_imports)]
```
soenkehahn, line 1: These `unused_imports` attributes (and the `dead_code` attribute below) are to silence warnings for rustc-versions `< 1.46`. Is there a better way to do this?

tests/panic_source_locations.rs
```diff
@@ -0,0 +1,39 @@
...
+    let f = || run_output!("false");
+    let panic_location = get_panic_location(f);
+    assert_eq!(
+        Some("tests/panic_source_locations.rs:19:16".to_string()),
+        panic_location
+    );
+}
+
+#[allow(dead_code)]
+fn get_panic_location(f: fn()) -> Option<String> {
```
soenkehahn, line 28: I think it'd make more sense if this function returned `Option<Location>` instead of `Option<String>`. But I couldn't make the lifetimes work. @casey: Are you up for a little lifetime puzzle? :) (I also think it's ok like this.)

tests/panic_source_locations.rs
```diff
@@ -0,0 +1,39 @@
...
+    assert_eq!(
+        Some("tests/panic_source_locations.rs:13:16".to_string()),
+        panic_location
+    );
+    let f = || run_output!("false");
+    let panic_location = get_panic_location(f);
+    assert_eq!(
+        Some("tests/panic_source_locations.rs:19:16".to_string()),
+        panic_location
+    );
```
soenkehahn, line 24: These two assertions are in the same test to make sure they're not executed in parallel. `get_panic_location` is not thread-safe, since it modifies the global panic handler. There's other ways to solve this, not sure they're better though.

tests/panic_source_locations.rs
casey: Can you make the imports conditional?

```rust
#[rustversion::since(1.46)]
use foo;
```

tests/panic_source_locations.rs
casey: I think `String` is okay too, which is convenient, because I don't think it's possible to solve the lifetime issues.

tests/panic_source_locations.rs
casey: Can't this also race with other tests, if those tests panic?

PR #210, opened 2022-05-04
Migrate tests to use python scripts instead of rust test helper

soenkehahn:


src/lib.rs
```diff
@@ -850,7 +861,13 @@ mod tests {
 
         #[test]
         fn capture_stderr() {
-            let Stderr(stderr) = run_output!(test_helper(), "write to stderr");
+            let script = TestScript::new(
+                r#"
+                    import sys
+                    print("foo", file=sys.stderr)
+                "#,
```
casey, line 868: If you use single quotes you can avoid raw strings:

```suggestion
                "
                    import sys
                    print('foo', file=sys.stderr)
                ",
```

casey:
Having the code be out-of-line doesn't bother me so much, so I think I still prefer keeping it in Rust vs adding Python. That being said, I don't feel super strongly about it. I can definitely see the readability benefit of having the code be in-line. Overall the PR looks good.

PR #211, opened 2022-05-05
Move context-integration-tests into their own sub-package

soenkehahn:
This PR moves the integration tests for the `Context` from an executable in the main package to its own sub-package. It now works very similar to the existing `memory-tests` sub-package. This reduces the complexity of the main `Cargo.toml` file.


casey:
This kind of seems like a wash to me, the main Cargo.toml gets less complicated, but now there's an extra binary 🤷‍♀️

soenkehahn:
> This kind of seems like a wash to me, the main Cargo.toml gets less complicated, but now there's an extra binary woman_shrugging

Yeah, it's not super obvious how this is better. :| Btw, there's not an extra binary now, the binary existed before in the main package. It just moved to a new sub-package.

This is in preparation for #210. In that PR I had to change `context_integration_tests.rs` and that incurred more dependencies in the `Cargo.toml` file that all had to be specified like [this](https://github.com/soenkehahn/cradle/blob/378e25338fc33f6c0b8e030e8b811c3277e0e145/Cargo.toml#L46..L59). Which I think is pretty bad, because it's kind of obfuscated that these are test dependencies. And then I thought it'd be good to make smaller changes, that's why I opened this PR.

casey:
Ah, gotcha. Yeah in that case it seems reasonable.

PR #213, opened 2022-07-30
Do not run system-dependent examples

jeandudey:
Hello!

I'm packaging _just_ on GNU Guix and found some friction when packaging this crate<sup>[1]</sup> (Guix needs all dependencies to be compiled an tested) because it requires access to some executables that are system dependent, so I've marked some of the doc-tests to not run (but will be compiled) so testing can be done on GNU Guix (each build is a container with minimal utilities needed to compile the source code with the additional dependencies).

[1]: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=56663

Also I've relaxed the version requirements on `Cargo.toml`, this to make maintenance easier on GNU Guix side as some of the dependencies might not be on the latest released version. (crates are updated but it's a slow process and maybe some crates have semver breaking changes). To overcome the MSRV issue on the CI, I pinned the `bitflags` crate on the GitHub workflow.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -27,7 +27,7 @@ pub enum Error {
     /// A `file not found` error occurred while trying to spawn
     /// the child process:
     ///
-    /// ```
+    /// ```no_run
     /// use cradle::prelude::*;
     ///
     /// let result: Result<(), Error> = run_result!("does-not-exist");
@@ -61,7 +61,7 @@ pub enum Error {
     CommandIoError { message: String, source: io::Error },
     /// The child process exited with a non-zero exit code.
     ///
-    /// ```
+    /// ```no_run
     /// use cradle::prelude::*;
     ///
     /// let result: Result<(), cradle::Error> = run_result!("false");
--- a/src/input.rs
+++ b/src/input.rs
@@ -20,7 +20,7 @@ use std::{
 /// For example you can pass in an executable as a String,
 /// and a variable number of arguments as a [`Vec`]:
 ///
-/// ```
+/// ```no_run
 /// use cradle::prelude::*;
 ///
 /// let executable = "echo";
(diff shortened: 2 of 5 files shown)


soenkehahn (changes requested):
Hi @jeandudey, thanks for the PR!

`cradle` relies on the doctests being run for test-coverage. So I don't think it's a good idea to disable them. Also, you say that guix *needs* all tests to be run. That also sounds like you shouldn't just disable them. :) Could you explain, which executables the test-suite relies on, that aren't available on the guix test environment when building `cradle`? I'm not a guix expert, but I would assume that there's some way to declare dependencies that are *only* needed to run the tests. Is that not the case?

Regarding the `bitflags` change: Could you explain what problem you ran into exactly? Which compiler version or versions did you use? And which `bitflags` versions are available?

Also, I think it could be helpful to split these two changes up into separate PRs. So that we can discuss them separately and not have one depend on the other.

jeandudey:
>Could you explain, which executables the test-suite relies on, that aren't available on the guix test environment when building cradle? I'm not a guix expert, but I would assume that there's some way to declare dependencies that are only needed to run the tests. Is that not the case?

Yeah, that's what I did with with it on Guix, declaring all the dependencies, [see here for the dependencies](https://issues.guix.gnu.org/56663#3-lineno85), xxd, coreutils and which. One of the doc tests also tries to access `$HOME` which is not available on the build container. If these tests test the functionality of the crate directly it would be better to switch them to normal tests (`#[test]`) so if they fail on Guix (or other distributions that package rust crates like Ubuntu or Debian) because of differences with a normal Linux distribution they can be skipped/ignored at the test phase.

>Regarding the bitflags change: Could you explain what problem you ran into exactly? Which compiler version or versions did you use? And which bitflags versions are available?

The code currently expects the exact version of `bitflags`, Guix only packages a single major version of the package (unless the package breaks semver and a specific minor one needs to be included) to make things easier, and, currently a higher version is packaged, so, build breaks.

The `bitflags` 1.2.1 crate can still be pinned on CI to avoid the MSRV compilation error.

That's also the reason for relaxing the other versions as Guix packages older o newer crates. For example, [see this](https://issues.guix.gnu.org/56663#3-lineno46).

>Also, I think it could be helpful to split these two changes up into separate PRs. So that we can discuss them separately and not have one depend on the other.

Yeah for sure!

PR #218, opened 2024-02-07
Add Output implementation for deserializing JSON

casey:
I figured that cradle was probably lonely from not having a PR for a while, so I decided to help.

I'm running a bunch of commands (`ord` and `bitcoin-cli`) which return JSON on standard output.

This PR adds a `Json` struct, which wraps any type that implements deserialize, and deserializes the command's standard output into an instance of that type. It's kind of unfortunate to have an output type which is specific to JSON, and not something that can be used to deserialize anything, but as far as I know, serde doesn't provide a trait which allows you to abstract over serialization formats, so I think adding a struct for each serialization format is the way.

`serde` and `serde_json` are pretty heavy dependencies, and `cradle` currently has no dependencies other than `rustversion`, so it might be wise to put this behind an optional, default-off feature flag.

(You will notice that there are no tests, and I can thus hear you clutching your pearls, but I promise to write some if you think this is a good idea!)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,13 +11,17 @@ keywords = ["child", "child-process", "command", "process", "shell"]
 categories = ["filesystem", "os"]
 exclude = ["/.github"]
 
+[features]
+serde = ["dep:serde", "dep:serde_json"]
+test_executables = ["nix", "serde"]
+
 [workspace]
 members = [".", "context-integration-tests", "memory-tests"]
 
 [dependencies]
 rustversion = "1.0.4"
-serde = "1.0.196"
-serde_json = "1.0.113"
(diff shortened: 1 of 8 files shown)


soenkehahn:
Hey, thanks for the PR! I think in general this would be very nice to have. And given that, for all I know, you're the only active user of cradle,  I think we should definitely add this!

I agree that it's a very heavy dependency and should be behind a feature flag, if that works for you. (Would the documentation on `docs.rs` be built for all available features, or just the default features?)

And yes, some tests would be nice. Clutch, clutch!

casey:
Closing, since I'm sadly no longer using cradle with JSON T_T

PR #220, opened 2025-10-04
Link to the Haskell version of cradle

sellout:
I was looking for the Haskell library, but (because I knew @soenkehahn was the author) I found this one instead and thought “oh, this is the name of the Rust crate, I must have been thinking of something else”.

I think having the same name across languages is good, but then they should also cross-link so you can quickly get to the right one from the other (I’m opening a similar PR against the Haskell package).


soenkehahn (approved):
Nice, thanks!

